PR #7, opened 2021-02-18
feat: support automatic json body for post requests

### LINKED ISSUE #4: Supporting JSON body
It would be great for POST request to support body as an object.
If this is the case, ohmyfetch will add the correct header and stringify the body.

danielroe:
closes #4

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -26,24 +26,20 @@ export interface $Fetch {
 }
 
 export function setHeader (options: FetchOptions, _key: string, value: string) {
-  const key = _key.toLowerCase()
+  let key = _key.toLowerCase()
   options.headers = options.headers || {}
-  if (Array.isArray(options.headers)) {
+  if ('set' in options.headers) {
+    ;(options.headers as Headers).set(key, value)
+  } else if (Array.isArray(options.headers)) {
     const existingHeader = options.headers.find(([header]) => header.toLowerCase() === key)
     if (existingHeader) {
       existingHeader[1] = value
     } else {
       options.headers.push([key, value])
     }
-  } else if ('set' in options.headers) {
-    ;(options.headers as Headers).set(key, value)
   } else {
-    const existingHeader = Object.keys(options.headers).find(header => header.toLowerCase() === key)
-    if (existingHeader) {
-      options.headers[existingHeader] = value
-    } else {
-      options.headers[key] = value
-    }
+    key = Object.keys(options.headers).find(header => header.toLowerCase() === key) || key
+    options.headers[key] = value
   }
 }
 
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -25,6 +25,28 @@ export interface $Fetch {
   raw<T = any>(request: FetchRequest, opts?: FetchOptions): Promise<FetchResponse<T>>
 }
 
+export function setHeader (options: FetchOptions, _key: string, value: string) {
+  const key = _key.toLowerCase()
+  options.headers = options.headers || {}
+  if (Array.isArray(options.headers)) {
+    const existingHeader = options.headers.find(([header]) => header.toLowerCase() === key)
+    if (existingHeader) {
+      existingHeader[1] = value
+    } else {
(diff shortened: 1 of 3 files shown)


Atinux (approved):

src/fetch.ts
```diff
@@ -31,8 +32,21 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
...
+          opts.body = JSON.stringify(opts.body)
+          opts.headers = opts.headers || {}
+          if (Array.isArray(opts.headers)) {
+            opts.headers.push(['content-type', 'application/json'])
+          } else if ('set' in opts.headers) {
+            ;(opts.headers as Headers).set('content-type', 'application/json')
+          } else {
+            opts.headers['content-type'] = 'application/json'
+          }
+        } catch {}
```
pi0, line 46: try/catch scope is wider than necessary (i think purpose was to handle JSON.stringify issues) but i think for that we may either keep original error or fallback to opts.body

src/fetch.ts
```diff
@@ -31,8 +32,21 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
       if (opts.params) {
         request = withQuery(request, opts.params)
       }
+      if (opts.body && typeof opts.body === 'object' && ['patch', 'post', 'put'].includes(opts.method?.toLowerCase() || '')) {
```
pi0, line 35: Body can be also of type `FormData` (typeof === object). We should ensure proto is pure object

pi0, line 35: Also i would create an array outside for keeping methods that accept post

src/fetch.ts
```diff
@@ -31,8 +32,21 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
...
+        try {
+          opts.body = JSON.stringify(opts.body)
+          opts.headers = opts.headers || {}
+          if (Array.isArray(opts.headers)) {
+            opts.headers.push(['content-type', 'application/json'])
```
pi0, line 40: Note: In array format, we might add duplicate content-type.

src/fetch.ts
```diff
@@ -32,18 +34,21 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
...
-          if (Array.isArray(opts.headers)) {
-            opts.headers.push(['content-type', 'application/json'])
-          } else if ('set' in opts.headers) {
-            ;(opts.headers as Headers).set('content-type', 'application/json')
+      if (opts.body && opts.body.toString() === '[object Object]' && payloadMethods.includes(opts.method?.toLowerCase() || '')) {
+        opts.body = JSON.stringify(opts.body)
+        opts.headers = opts.headers || {}
+        if (Array.isArray(opts.headers)) {
+          const index = opts.headers.findIndex(([header]) => header.toLowerCase() === 'content-type')
+          if (index >= 0) {
```
pi0, line 42: - Can't we inline with `headers.find`?
- Lowercase check is not happening for object mode

With currently increased logic, i would refactor header setting to a utility function

src/fetch.ts
```diff
@@ -22,6 +25,28 @@ export interface $Fetch {
...
+    const existingHeader = options.headers.find(([header]) => header.toLowerCase() === key)
+    if (existingHeader) {
+      existingHeader[1] = value
+    } else {
+      options.headers.push([key, value])
+    }
+  } else if ('set' in options.headers) {
+    ;(options.headers as Headers).set(key, value)
+  } else {
+    const existingHeader = Object.keys(options.headers).find(header => header.toLowerCase() === key)
```
pi0, line 41: By early returning on `set in` case, we can dedup this line

pi0:
It think would be nice mentioning this feature in readme :)

src/fetch.ts
danielroe: not quite sure how to dedupe it - though it feels very close with the array/object syntax. how were you thinking?

pi0: By early return in first if case we can move variable to outer scope :)

Issue #13, opened 2021-09-15
Make parseoptional


pi0:
Context: https://twitter.com/intrnl0/status/1438179142512504836

Two ideas:
- Disable deserialization at all with `parse: false`
- Allow providing another deserializer with `parse: JSON.parse`

chrstnfrrs:
This looks like a great first issue, can I pick this up later today?

pi0:
Suee @chrstnfrrs

PR #14, opened 2021-09-16
Issue-13: Make parse optional

chrstnfrrs:
Closes: #13 

**Description**
This PR aims to make parsing optional.

- It creates a new object to pass through as options, setting parse to true and then spreading consumer options. This will overwrite the parse set to true, if parse is passed through.
- If parse is set, then it parses the JSON body and automatically adds headers. If parse is set to false, then body is maintained and headers are not changed.
- Parse is added to the FetchOptions type as an optional boolean.


chrstnfrrs:
I misunderstood what the Issue was requesting. This solution is incorrect and I will open a new :pr: with the correct solution.

PR #20, opened 2021-10-20
fix: check for `globalThis` before fallback to shims

ms-fadaei:
I just updated the getGlobal function to get the best result! Not a significant change at all 😄

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,13 +2,12 @@ import { createFetch } from './base'
 
 export * from './base'
 
-// ref: https://mathiasbynens.be/notes/globalthis
-const getGlobal = () => {
+// ref: https://github.com/tc39/proposal-global
+const getGlobal = function () {
   if (typeof globalThis !== 'undefined') { return globalThis }
   if (typeof self !== 'undefined') { return self }
   if (typeof window !== 'undefined') { return window }
   if (typeof global !== 'undefined') { return global }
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -2,12 +2,12 @@ import { createFetch } from './base'
 
 export * from './base'
 
-// ref: https://github.com/tc39/proposal-global
```
pi0, line 5: Can we please keep this? Or do you have a more updated ref?

src/index.ts
```diff
@@ -2,12 +2,12 @@ import { createFetch } from './base'
...
 const getGlobal = function () {
+  if (typeof globalThis !== 'undefined') { return globalThis }
   if (typeof self !== 'undefined') { return self }
   if (typeof window !== 'undefined') { return window }
   if (typeof global !== 'undefined') { return global }
-  throw new Error('unable to locate global object')
+  return (function() { return this })();
```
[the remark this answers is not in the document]
ms-fadaei: You right!

src/index.ts
pi0: Well we can mention with `this` fallback addition :p BTW we also need to update unenv https://github.com/unjs/unenv/blob/main/src/runtime/polyfill/globalThis.ts

ms-fadaei: I can update the `unenv` if you want and use this repository as a ref comment 😄

PR #25, opened 2021-11-08
fix: remove baseurl append on retry

### LINKED ISSUE #24: baseURL is being appended on every retry
**Describe the bug**
On a request that has both `baseURL` and `retry > 0`, baseURL will be appended on every retry attempt.

**To Reproduce**
Change the code in `playground/index.ts` to the following:

``` typescript
async function main () {
  // const r = await $fetch<string>('http://google.com/404')
  // const r = await $fetch<string>('http://httpstat.us/500')
  // const r = await $fetch<string>('http://httpstat/500')
  const r = await $fetch<string>('404', { baseURL: 'http://httpstat.us/', retry: 3 }).catch(err => err)
  // eslint-disable-next-line no-console
  console.log(r.request)
}
```

Then run the following command:
```
yarn play
```

Output will show as:
```
http://httpstat.us/http://httpstat.us/http://httpstat.us/http://httpstat.us/404
```

**Expected behavior**
Output should be the following:
```
http://httpstat.us/404
```

OliverVorasai:
Removes further baseURL appends when retrying.

closes #24

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -54,6 +54,7 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
       if (retries > 0) {
         return $fetchRaw(request, {
           ...opts,
+          baseURL: '',
           retry: retries - 1
         })
       }
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -1,5 +1,5 @@
 import destr from 'destr'
-import { joinURL, withQuery } from 'ufo'
+import { withBase, withQuery } from 'ufo'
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for PR @OliverVorasai. I guess we can use new `ufo.withBase` utility on [L71](https://github.com/unjs/ohmyfetch/blob/main/src/fetch.ts#L71)

OliverVorasai:
That is an excellent suggestion @pi0. I have made the change.

pi0 (approved):

PR #29, opened 2021-11-16
fix(fetch): use `application/json` for array body

### LINKED ISSUE #28: JSON headers is not working for array of objects
Hi there.
I was working with this library and I found out in a special case, it doesn't add header `content-type: application/json` to the request.

#### What I did
```js
$fetch("/api/endpoint", {method: "POST", body: [{num: 42}, {num: 43}]})
```
#### What I expected
Automatically add `content-type: application/json` header and stringify the body.

#### What actually happened
the `content-type: application/json` header was not added and the request body replaced with `[object Object],[object Object]` 

#### Details
The request was not contain `content-type: application/json` header because as this feature implemented [here](https://github.com/unjs/ohmyfetch/blob/4ab7022ae96c7990d2488eacc04c06d094d4de31/src/fetch.ts#L80), the `toString()` of the body in this case is not equal to `[object Object]`  (actually it is equal to `[object Object],[object Object]` in this case). So automatic json headers and body stringify didn't happen.

Narixius:
It may fix #28

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -77,7 +77,7 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
         request = withQuery(request, opts.params)
       }
       const hasPayload = payloadMethods.includes((opts.method || '').toLowerCase())
-      if (opts.body && opts.body.toString().includes('[object Object]') && hasPayload) {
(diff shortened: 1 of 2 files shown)


pi0 (approved):

pi0:
Thanks! Released as ohmyfetch@0.4.7

PR #36, opened 2021-12-10
fix(fetch): prevent body stringify when `FormData`

benjamincanac:


src/fetch.ts
```diff
@@ -77,7 +77,9 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
         request = withQuery(request, opts.params)
       }
       const hasPayload = payloadMethods.includes((opts.method || '').toLowerCase())
-      if (opts.body && typeof opts.body === 'object' && hasPayload) {
+      const supportsFormData = typeof FormData === 'function'
+      const isFormData = supportsFormData && opts.body instanceof FormData
```
pi0, line 81: We cannot rely on global FormData detection. One can directly use a package like [form-data](https://www.npmjs.com/package/form-data) in Node.js without global polyfill.

src/fetch.ts
```diff
@@ -77,7 +77,9 @@ export function createFetch ({ fetch }: CreateFetchOptions): $Fetch {
...
+      if (opts.body && typeof opts.body === 'object' && !isFormData && hasPayload) {
```
danielroe, line 82: how about something like this instead:
```suggestion
      if (opts.body && typeof opts.body === 'object' && hasPayload && opts.body.toString() !== '[object FormData]') {
```

pi0: Pushed a generic fix but seems a good idea for using `[object FormData]` to append proper content-type (adding in short)

pi0:
This issue should be fixed on 0.4.9

PR #39, opened 2021-12-14
feat: return blob if `content-type` isn't text, svg, xml or json

### LINKED ISSUE #38: Blob method does not exist in the result
How can I return the result in blob? It seems that .blob() does not exist

```
- Operating System: `Darwin`
- Node Version:     `v16.13.1`
- Nuxt Version:     `3.0.0-27319101.3e82f0f`
- Package Manager:  `npm@8.1.2`
- Bundler:          `Vite`
- User Config:      `css`, `watch`, `build`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

```js
const {data: banners, refresh: bannersFetcher} = await useAsyncData(
    'banners',
    async () => {
      const banners = await $fetch('/api/mc/banners')
        .then((data) => {
          return Promise.all(_.map(data, async (banner) => {
            const blob = await $fetch(banner.data).then(res => res.blob())
            console.log('Hey', blob)
          }))
        })
      // console.log(banners)
      return []
    }
)
```

danielroe:
resolves  #38

For binary content types, `$fetch` will instead return a `Blob` object.

You can optionally specify `blob`, `arrayBuffer` or `text` to force parsing the body with the respective `FetchResponse` method.

```js
// Use JSON.parse
await $fetch('/movie?lang=en', { parseResponse: JSON.parse })

// Return text as is
await $fetch('/movie?lang=en', { parseResponse: txt => txt })

// Get the blob version of the response
await $fetch('/api/generate-image', { responseType: 'blob' })
```

## Typing
```ts
$fetch('api') // Promise<any>
$fetch<boolean>('api') // Promise<boolean>
$fetch<boolean>('api', { responseType: 'text' }) // Type '"text"' is not assignable to type '"json" | undefined'.ts(2322)
$fetch('api', { responseType: 'json' }) // Promise<any>
$fetch<boolean>('api', { responseType: 'json' }) // Promise<boolean>
$fetch('api', { responseType: 'text' }) // Promise<string>
$fetch('api', { responseType: 'arrayBuffer' }) // Promise<ArrayBuffer>
$fetch('api', { responseType: 'blob' }) // Promise<Blob>
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -20,21 +20,27 @@ export function isJSONSerializable (val: any) {
   return val.constructor?.name === 'Object' || typeof val.toJSON === 'function'
 }
 
-// This provides reasonable defaults for the correct parser based on Content-Type header
-export function detectContentMethod (contentType = ''): 'text' | 'blob' {
-  // Value might look like: `application/json; charset=utf-8`
-  switch (true) {
-    case !contentType:
-    case contentType.startsWith('text/'):
-    case contentType.startsWith('image/svg'):
-    case contentType.startsWith('application/xml'):
-    case contentType.startsWith('application/xhtml'):
-    case contentType.startsWith('application/html'):
-    case contentType.startsWith('application/json'):
-    case contentType.startsWith('application/ld+json'):
-      return 'text'
+const textTypes = new Set([
+  'image/svg',
+  'application/xml',
+  'application/xhtml',
+  'application/html',
+  'application/json',
+  'application/ld+json'
+])
 
-    default:
-      return 'blob'
+// This provides reasonable defaults for the correct parser based on Content-Type header
+export function detectContentMethod (_contentType = ''): 'text' | 'blob' | 'json' | 'arrayBuffer' {
+  if (!_contentType) {
+    return 'text'
   }
+
+  // Value might look like: `application/json; charset=utf-8`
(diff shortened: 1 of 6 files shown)


test/index.test.mjs
```diff
@@ -4,6 +4,7 @@ import { createApp, useBody } from 'h3'
 import { expect } from 'chai'
 import { Headers } from 'node-fetch'
 import { $fetch } from 'ohmyfetch'
+import { Blob } from 'fetch-blob'
```
pi0, line 7: Implicit dependency

src/utils.ts
```diff
@@ -19,3 +19,32 @@ export function isJSONSerializable (val: any) {
...
+  'image/svg',
+  'application/xml',
+  'application/xhtml',
+  'application/html'
+])
+
+const jsonTypes = new Set(['application/json', 'application/ld+json'])
+
+// This provides reasonable defaults for the correct parser based on Content-Type header.
+export function detectContentMethod (_contentType = ''): 'text' | 'blob' | 'json' | 'arrayBuffer' {
```
pi0, line 33: I would make a type for return type to reuse with `ResponseType `

pi0 (approved):

pi0 (approved):

PR #41, opened 2021-12-21
feat: allow user overrides of `content-type` and `accept`

### LINKED ISSUE #40: Disable Accept and Content-Type header rewriting if they are already defined
Hi, i need send JSON body via **PATCH** method with `'Content-Type': 'application/merge-patch+json'` head but OhMyFetch always rewrite it to  `application/json`. 
I would suggest that if the user defines a "Accept" or "Content-type" header, OhMyFetch will not overwrite them.

danielroe:
resolves #40 

This PR does three things:
* it normalizes headers to an object accepting get/set methods (internally), rather than on each get/set call, which should improve performance slightly
* it exposes a `getHeader` function in pair to `setHeader`
* it allows users to override `content-type` and `accept` headers

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -29,8 +29,6 @@ export interface $Fetch {
 
 export function normalizeHeaders (_opts: FetchOptions): Pick<Headers, 'get' | 'set' | 'has'> {
   const opts = (_opts as FetchOptions & { _headers: Pick<Headers, 'get' | 'set' | 'has'> })
-
-  // @ts-ignore
   if (opts._headers) { return opts._headers }
 
   opts.headers = opts.headers || {}
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -40,7 +40,7 @@ export function normalizeHeaders (options: FetchOptions): Pick<Headers, 'get' | 
     return {
       get: (_key: string) => {
         const key = _key.toLowerCase()
-        return findHeader(_key)?.[1] || null
+        return findHeader(key)?.[1] || null
       },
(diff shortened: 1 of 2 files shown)


src/fetch.ts
```diff
@@ -27,24 +27,47 @@ export interface $Fetch {
...
+          existingHeader[1] = value
+        } else {
+          headers.push([key, value])
+        }
+      }
+    }
+  }
+
+  const headers = options.headers
+  const findHeaderKey = (key: string) => Object.keys(headers).find(header => header.toLowerCase() === key)
```
pi0, line 53: Perf: We could normalizer either Array into Object or Object into array entities to merge logic and reduce bundle overhead. I would suggest using Object as normalize base since gives us direct lookup possibility

src/fetch.ts
```diff
@@ -27,24 +27,35 @@ export interface $Fetch {
...
+  if (Array.isArray(opts.headers)) {
+    headers = Object.fromEntries(opts.headers.map(([key, value]) => [key.toLowerCase(), value]))
   } else {
-    const existingHeader = Object.keys(options.headers).find(header => header.toLowerCase() === key)
-    options.headers[existingHeader || key] = value
+    headers = Object.fromEntries(Object.entries(opts.headers).map(([key, value]) => [key.toLowerCase(), value]))
   }
+
+  opts.headers = headers
+  opts._headers = {
```
pi0, line 45: Mutating user options is probably not the best idea as it can affect next calls if they use shared reference and implicitly mutate headers in options (normalized will keep referencing to the old one)

PR #49, opened 2022-01-18
fix: use `_data` rather than `data` to store deserialized response

danielroe:
resolves https://github.com/nuxt/framework/issues/2768


pi0:
Thanks!

PR #55, opened 2022-03-01
fix: expose $fetch.create in CJS Proxy

### LINKED ISSUE #44: Error using $fetch.create using CJS modules.
I was working in a Nuxt2 app using `ohmyfetch` in the background with a home-made module, but the Nuxt2 app displays me that the`$fetch` export has not the method `create`, so I investigate and I found that the CJS module (what it was imported by Webpack after the bundle process) it not exports the method.

## Steps to reproduce the error
1. Create a new JS proyect with the `ohmyfetch` dependency.
2. Create a `index.js` file with the following content.
```js
const { $fetch } = require('ohmyfetch')

const scopedFetch = $fetch.create({});
```
3. Try to run the file.

## What I was expected
It should not thrown an error.

## Error description
```
const scopedFetch = $fetch.create({});
                           ^

TypeError: $fetch.create is not a function
    at Object.<anonymous> (/home/bantonio/PhpstormProjects/test-ohmyfetch/index.js:3:28)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
```

P4sca1:
This allows to use $fetch.create in CJS environments. A caveat is that the function is not synchronous, but returns a promise.
See #44

### DIFF
--- a/cjs/index.cjs
+++ b/cjs/index.cjs
@@ -4,3 +4,4 @@ const createCaller = name => (input, init) => getExport(name).then(fn => fn(inpu
 exports.fetch = createCaller('fetch')
 exports.$fetch = createCaller('$fetch')
 exports.$fetch.raw = (input, init) => getExport('$fetch').then($fetch => $fetch.raw(input, init))
+exports.$fetch.create = (opts) => getExport('$fetch').then($fetch => $fetch.create(opts))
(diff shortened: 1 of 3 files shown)


danielroe (approved):

pi0:
Thanks for PR @P4sca1 but I'm afraid this is not a good solution since as you mentioned, makes behavior different by returning a Promise. We could return a deferred function instead that resolves promise on first `$fetch` or `$fetch.raw` call.

P4sca1:
I used the patch as a workaround so that `$fetch.create` works in a Nuxt.js plugin where the fetch instance is only used client-side. But I understand that it is not a good solution to solve the problem.

PR #69, opened 2022-04-16
feat: `createClient` REST client using JS proxies

johannschopplich:
Hi there, @pi0, so here it is as discussed!

## Minimal, Type-Safe REST Client

[`unrested`](https://github.com/johannschopplich/unrested) is now part of `ofetch` and has been cleaned up internally. Test coverage has been increased as well and updated to use `listhen` just like the tests for `$fetch`.

This package now ships with `createClient`, a tiny REST client built upon JS proxies. It re-uses every UnJS gem, like `ufo` and `defu`.

The [documentation has been added](https://github.com/johannschopplich/ohmyfetch/tree/feat/rest-client#-rest-client-with-createclient) as well.

I hope this PR looks good to you. If I missed something, let me know. Thanks!

## Notes

- ~~I removed `@types/node-fetch` from this repo. `node-fetch` provides its own type definitions, so we don't need this installed.~~ Already done upstream.
- The build config's option `emitCJS: false` is deprecated in the latest `unbuild` version. The package's `exports` declare CJS build artifacts, so I enabled them again. 
- Some minor refactorings have been applied to the code base.
- Heading capitalization of the main README has been made consistent.

## Features

You can use `ofetch` as a minimal, type-safe REST client for any given API. The API builder uses JS proxies under the hood.

```js
import { createClient } from 'ofetch'

const api = createClient({
  // Set optional defaults for `$fetch`
  baseURL: '<baseURL>'
})
```

The `baseURL` will default to `/` if not set.

### Path Segment Chaining

Chain single path segments or path ids by a dot. You can even type the response of your request!

```js
// GET request to <baseURL>/users
const users = await api.users.get<UserResponse>()

// For GET request you can add search params
// <baseURL>/users?search=john
const users = await api.users.get<UserResponse>({ search: 'john' })

// Chains can get as deep as you wish
// <baseURL>/deeply/nested
const data = await api.deeply.nested.get()
```

To include dynamic API path segments, you have two options:

```js
// Typed GET request to <baseURL>/users/1
const userId = 1
// … using the chain syntax:
const user = await api.users(userId).get<UserResponse>()
// … or the bracket syntax:
const user = await api.users[`${userId}`].get<UserResponse>()
```

### HTTP Request Methods

Add the appropriate method to the end of your API call. The following methods are supported:

- `get()`
- `post()`
- `put()`
- `delete()`
- `patch()`

### Payload Requests

For HTTP request methods supporting a payload, add it to the method call:

```js
// POST request to <baseURL>/users
const response = await api.users.post({ name: 'foo' })
```

### Overwrite Default Options

You can add/overwrite `$fetch` options on a method-level as well:

```js
const response = await api.users.get({
  headers: {
    'Cache-Control': 'no-cache'
  }
})
```

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -41,16 +41,10 @@ export function createClient<R extends ResponseType = 'json'> (
           data?: RequestInit['body'] | Record<string, any>,
           opts: FetchOptions<R> = {}
         ) => {
-          switch (method) {
-            case 'GET':
-              if (data) {
-                url = withQuery(url, data as QueryObject)
-              }
-              break
-            case 'POST':
-            case 'PUT':
-            case 'PATCH':
-              opts.body = data
+          if (method === 'GET' && data) {
+            url = withQuery(url, data as QueryObject)
+          } else if (['POST', 'PUT', 'PATCH'].includes(method)) {
+            opts.body = data
           }
 
           opts.method = method
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -1,0 +1,68 @@
+import { resolveURL, withQuery } from 'ufo'
+import type { QueryObject } from 'ufo'
+import type { FetchOptions } from './fetch'
+import type { ResponseType, MappedType } from './types'
+import { $fetch } from './undici'
+
+export type ClientMethodHandler = <T = any, R extends ResponseType = 'json'>(
+  data?: RequestInit['body'] | Record<string, any>,
+  opts?: Omit<FetchOptions<R>, 'baseURL' | 'method'>
+) => Promise<MappedType<R, T>>
+
+export type ClientBuilder = {
+  [key: string]: ClientBuilder;
+  (...segmentsOrIds: (string | number)[]): ClientBuilder
+} & {
+  get: ClientMethodHandler
+  post: ClientMethodHandler
+  put: ClientMethodHandler
+  delete: ClientMethodHandler
+  patch: ClientMethodHandler
+}
+
+export function createClient<R extends ResponseType = 'json'> (
+  url: string,
+  defaults: Omit<FetchOptions<R>, 'method'> = {}
+): ClientBuilder {
+  // Callable internal target required to use `apply` on it
+  const internalTarget = (() => {}) as ClientBuilder
+
+  const p = (url: string): ClientBuilder =>
+    new Proxy(internalTarget, {
+      get (_target, key: string) {
+        const method = key.toUpperCase()
+
+        if (!['GET', 'POST', 'PUT', 'DELETE', 'PATCH'].includes(method)) {
+          return p(resolveURL(url, key))
+        }
+
+        const handler: ClientMethodHandler = <T = any, R extends ResponseType = 'json'>(
+          data?: RequestInit['body'] | Record<string, any>,
+          opts: Omit<FetchOptions<R>, 'baseURL' | 'method'> = {}
+        ) => {
+          switch (method) {
+            case 'GET':
+              if (data) { url = withQuery(url, data as QueryObject) }
+              break
+            case 'POST':
+            case 'PUT':
+            case 'PATCH':
+              opts.body = data
+          }
+
+          return $fetch<T, R>(url, {
+            ...(defaults as unknown as FetchOptions<R>),
+            ...opts,
+            method
+          })
+        }
+
+        return handler
+      },
+      apply (_target, _thisArg, args: (string | number)[] = []) {
+        return p(resolveURL(url, ...args.map(i => `${i}`)))
+      }
+    })
+
+  return p(url)
+}
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -50,8 +50,8 @@ export function createClient<R extends ResponseType = "json"> (
             ...(globalOptions as FetchOptions<R>),
             ...options,
             headers: {
-              ...headersToObject(globalOptions.headers || {}),
-              ...headersToObject(options.headers || {})
+              ...headersToObject(globalOptions.headers),
+              ...headersToObject(options.headers)
             },
             method
           });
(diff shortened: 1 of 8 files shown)


build.config.ts
```diff
@@ -2,10 +2,12 @@ import { defineBuildConfig } from 'unbuild'
...
   declaration: true,
-  emitCJS: false,
   entries: [
     'src/index',
     'src/node',
     'src/undici'
-  ]
+  ],
+  rollup: {
+    emitCJS: true
```
pi0, line 11: CJS support is limited and via a manual entry (https://github.com/unjs/ohmyfetch/tree/main/cjs). We cannot have `require('node-fetch')`. You can drop both options.

src/builder.ts
```diff
@@ -0,0 +1,70 @@
...
+} & {
+  get: ClientMethodHandler
+  post: ClientMethodHandler
+  put: ClientMethodHandler
+  delete: ClientMethodHandler
+  patch: ClientMethodHandler
+}
+
+export function createClient<R extends ResponseType = 'json'> (
+  url: string,
```
pi0, line 25: We can simplify interface to `createClient({ baseURL, ...options })`.

src/builder.ts
```diff
@@ -0,0 +1,70 @@
...
+          return p(resolveURL(url, key))
+        }
+
+        const handler: ClientMethodHandler = <T = any, R extends ResponseType = 'json'>(
+          data?: RequestInit['body'] | Record<string, any>,
+          opts: FetchOptions<R> = {}
+        ) => {
+          switch (method) {
+            case 'GET':
+              if (data) {
```
pi0, line 46: Code-style: I would use if/else for this.

src/builder.ts
```diff
@@ -0,0 +1,70 @@
...
+              break
+            case 'POST':
+            case 'PUT':
+            case 'PATCH':
+              opts.body = data
+          }
+
+          opts.method = method
+
+          return $fetch<T, R>(url, defu(opts, defaults) as FetchOptions<R>)
```
pi0, line 58: As long as options are flat, we probably don't need `defu` and can use `{ ...defaults, ...opts }`

johannschopplich: That was my initial idea as well. But what about extending `headers`? We could extend them on a request method level with defu.

build.config.ts
johannschopplich: I see. Thanks for the explanation.

src/builder.ts
johannschopplich: Great idea. So you mean `/` as the default base URL, right?
Please take a look at the current implementation: https://github.com/unjs/ohmyfetch/pull/69/commits/79098d08b2a4dcd106f6d52c988ea7881abb1525

src/builder.ts
pi0: Good point about headers! One issue with header merging is that the user can potentially pass a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) class as well and well that
 won't be merged as expected (until we [normalize `headers`](https://github.com/unjs/ohmyfetch/blob/26641a4405826de6281e357bdd665df692d0f340/src/fetch.ts#L110)). What about introducing new `defaultHeaders` global option?

johannschopplich: Right, but wouldn't a new option like `defaultHeaders` have to merged as well? I think normalizing headers is the more semantic approach, rather than introducing a new option. Let's postpone that and don't extend the scope of this PR too much. I think another PR focusing on normalizing headers would be cleaner. 🙂 

I have dropped defu and will introduce header merging later. Would do you say?

johannschopplich:
Rebased on `main`.

johannschopplich:
New feature: normalize headers and merge global with local headers.

For SSR compatibility, we check if `Headers` is defined. 

I know you're busy on Nuxt 3, Docus etc. So maybe this feature can still make in into `ofetch`. 🙂

src/builder.ts
johannschopplich: Merging headers is supported since https://github.com/unjs/ofetch/pull/69/commits/41e267ec4b1b02d072053078e5bd4db203f0a091.

johannschopplich:
@pi0 Rebased on `main` since the rename to `ofetch.` 🙂

therealokoro:
I upvote this. Looks awesome

miguelrk:
+1 for merging this! 🙏🏼

johannschopplich:
Hey there, @pi0, I have rebased and update the code to match the latest changes, including formatting with Prettier. I'd love you to find time for PR, but no rush. It's enduring. 🙂

pi0:
For everyone, we are working with @johannschopplich to introduce it as a new standalone unjs project ⚡

stychu:
> For everyone, we are working with @johannschopplich to introduce it as a new standalone unjs project ⚡

Thats nice! Where we can track the progress? 
Also I would like to request if possible to add an `schema adapter` layer for such so u could define your schema validators for the input,output,body etc. if possible. Sommething like [react-hook-form](https://react-hook-form.com/get-started#SchemaValidation) has

johannschopplich:
@stychu Thanks for the suggestion! I will share it once the library is ready. Let's open issues over at the new repo. 🙂

HigherOrderLogic:
Hi, any update on the new package [`api-party`](https://www.npmjs.com/package/api-party)?

johannschopplich:
@HigherOrderLogic Good question! `api-party` may take some time as it is still unclear how to proceed with the library. It should match the intentions of the UnJS ecosystem. In its current state, it may not.

However, I have released [`apiverse`](https://github.com/johannschopplich/apiverse), which _is_ a version of `api-party` and lets me test the idea in real-world scenarios. Feel free to use it and provide feedback until `api-party` is ready. 🙂

PR #74, opened 2022-04-27
Added FormData support to body

ozancakir:
Stringify FormData makes sending FormData post imposible. Applied an easy fix by checking body types


ozancakir:
Failed couple of test, witdrawing PR

PR #80, opened 2022-05-10
fix: only serialize JSON bodies

### LINKED ISSUE #58: [BUG] About send a request with body text type
https://github.com/unjs/ohmyfetch/blob/5307000f12ad2414e07f194bb049d78cef654cd4/src/fetch.ts#L108-L109
If I want to send a text body post request, it will `JSON.stringify` my text and add `"` before and after my text.

image

P4sca1:
Previously, everything that looked like it could be serialized, was actually serialized, no matter what content type was set.
This includes arbitrary strings. When you tried to send a string like `key "myval"` to the server using ohmyfetch, the server received `"key \\"myval\\""`.

Closes #58

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -105,22 +105,21 @@ export function createFetch (globalOptions: CreateFetchOptions): $Fetch {
         ctx.request = withQuery(ctx.request, ctx.options.params)
       }
       if (ctx.options.body && isPayloadMethod(ctx.options.method)) {
-        ctx.options.headers = new Headers(ctx.options.headers)
-
-        // Default to application/json for content-type and accept,
-        // when not explicitly set, and the body is JSON serializable.
         if (isJSONSerializable(ctx.options.body)) {
+          // Automatically JSON stringify request bodies, when not already a string.
+          ctx.options.body = typeof ctx.options.body === 'string'
+            ? ctx.options.body
+            : JSON.stringify(ctx.options.body)
+
+          // Set Content-Type and Accept headers to application/json by default
+          // for JSON serializable request bodies.
+          ctx.options.headers = new Headers(ctx.options.headers)
           if (!ctx.options.headers.has('content-type')) {
             ctx.options.headers.set('content-type', 'application/json')
           }
           if (!ctx.options.headers.has('accept')) {
             ctx.options.headers.set('accept', 'application/json')
           }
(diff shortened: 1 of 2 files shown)


src/fetch.ts
```diff
@@ -105,16 +105,23 @@ export function createFetch (globalOptions: CreateFetchOptions): $Fetch {
         ctx.request = withQuery(ctx.request, ctx.options.params)
       }
       if (ctx.options.body && isPayloadMethod(ctx.options.method)) {
+        ctx.options.headers = new Headers(ctx.options.headers)
+
+        // Default to application/json for content-type and accept,
+        // when not explicitly set, and the body is JSON serializable.
         if (isJSONSerializable(ctx.options.body)) {
-          ctx.options.body = JSON.stringify(ctx.options.body)
```
pi0, line 109: Can't we keep check inside this branch?

P4sca1: Separating the logic to set default headers, from the logic to automatically stringify JSON bodies, makes the code more readable in my opinion.

pi0: We could directly stringify here but make an exception for type: string. And remove header checking logic.

P4sca1: Is there any reason to stringify request bodies that have a different content type than application/json?

From my experience, implicitly stringifying leads to more issues than it solves.

pi0: Probably not. But then what would `options.body` value that is both serializable and we don't intend to serialize?

P4sca1: For example when you want to send raw strings or numbers that need to be send unmodified to the server.

pi0: Yes strings should be fixed as is. That's for the linked issue. JSON.stringify on other primitives does nothing. `JSON.stringify(2)` is `2` for instance.

P4sca1: So, instead of my proposed changes, you want to adjust the `isJSONSerializable` function to return `false` for strings? Is that correct?

pi0: I would still add application/json as default content type for string body if not specified. We could add a simple inline check like `typeof s === 'string' ? s : JSON.stringify(s)`

pi0 (approved):

pi0:
Thanks!

PR #97, opened 2022-06-22
Added the missing create function

m0dch3n:
Fix: https://github.com/unjs/ohmyfetch/issues/96


m0dch3n:
https://github.com/unjs/ohmyfetch/issues/44

Seems this makes create a promise

Issue #99, opened 2022-06-23
response.body.getReader error


okxiaoliang4:
when using `$fetch.raw` the `getReader()` will throw a error

example: 
```ts
const response = await $fetch.raw(chunk.url)
response.body.getReader()
```

```shell
TypeError: Failed to execute 'getReader' on 'ReadableStream': ReadableStreamDefaultReader constructor can only accept readable streams that are not yet locked to a reader
```

okxiaoliang4:
I think the error is here.
https://github.com/unjs/ohmyfetch/blob/9a94311432161766204f82615f7a98406e922212/src/fetch.ts#L135-L146

Wellan-Arthur:
I get that [response.data is deprecated](https://github.com/node-fetch/node-fetch/issues/1000), but my current workaround for this specific issue is `response._data`.
Hope this helps for quick fixes !

PR #100, opened 2022-06-23
feat: support `responseType: 'stream'` as `ReadableStream`

okxiaoliang4:
close: #99


pi0 (approved):
Nice PR! I've added a small change to automatically use stream with `application/octet-stream` content-type

PR #104, opened 2022-07-06
chore: refactor some defaults using `unjs/defu`

huynl-96:
Using `unjs/defu` to assign default properties recursively


pi0:
Thanks for the PR but i avoided to do this saving few more bytes from bundle overhead since defaults could be easily extended manually. We can reconsider this if had more complicated cases in the future.

PR #105, opened 2022-07-06
refactor: add `query` alias to `params` option

### LINKED ISSUE #73: rename `params` to `query`
`params` can have different meanings in different contexts. `fetch(URL, { query }` is probably more readable.

huynl-96:
Resolves #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ By using `baseURL` option, `$fetch` prepends it with respecting to trailing/lead
 await $fetch('/config', { baseURL })
 ```
 
-## ✔️ Adding query
+## ✔️ Adding Query Params
 
 By using `query` option, `$fetch` adds search query to URL by preserving query in request itself using [ufo](https://github.com/unjs/ufo):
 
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for PR @huynl-96. I think we can introduce query and deprecate params without breaking changes and it is good to go.

huynl-96:
@pi0 I've added a deprecated warning for `params`. Thanks for your advice

pi0:
Hi @huynl-96. Sorry it took long for me back to this. I had been thinking it might be better to support both options as alias (without deprecation) because the same "query string" is also named "URL search params" in other standards. Also using "Query search params" in docs to make it readable in both ways.

huynl-96:
@pi0 
> Sorry it took long for me back to this 

Don't worry about it, hope you are feeling better now.
It's good to finally have you back with us

> I had been thinking it might be better to support both options as alias (without deprecation) because the same "query string" is also named "URL search params" in other standards. Also using "Query search params" in docs to make it readable in both ways.

Looks good to me, no concern about it

Issue #111, opened 2022-07-28
Retry is also triggered when it is stopped manually


amihhs:
Retry is also triggered when it is stopped manually, 
if I make a global automatic stop repeat request, this will result in loop
```js
const peddlingReq = new Map()
const defaultOption = {
  onRequest: (ctx) => {
    if (peddlingReq.has(ctx.url))
      peddlingReq.get(ctx.url).abort()
    else {
      const controller = new AbortController()
      peddlingReq.set(ctx.url, controller)
      ctx.options = { ...ctx.options, signal: controller.signal }
    }
  }
}
function apiFetch(url, options) {
  const _opt = Object.assign({}, defaultOption, options)
  return $fetch(url, _opt)
}

function init() {
  apiFetch('/a', { retry: 3 })
  apiFetch('/a', { retry: 3 })
}

init()
```
 debug [L70](https://github.com/unjs/ohmyfetch/blob/9fed5b45aafec9296ad30899438a22ddb84b3886/src/fetch.ts#L64) can see multiple execution

PR #112, opened 2022-07-28
fix: do not  retry when fetch is aborted

amihhs:
Stop automatic retry by judging the wrong name【AbortError】、
https://developer.mozilla.org/en-US/docs/Web/API/DOMException#error_names


amihhs:
> [#111](https://github.com/unjs/ohmyfetch/issues/111)

pi0:
Thanks!

PR #113, opened 2022-07-29
fix: do not add baseURL to urls with protocol

danielroe:
Currently:

```js
$fetch('https://test.com', { baseURL: 'https://foo.com' })
```

will make a request to `https://test.com/https://foo.com`.

We could also change the behaviour of `withBase` to do nothing if passed an absolute URL, but I think it makes more sense within `ohmyfetch`.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -1,5 +1,5 @@
 import destr from 'destr'
-import { withBase, withQuery } from 'ufo'
+import { withBase, withQuery, hasProtocol } from 'ufo'
 import type { Fetch, RequestInfo, RequestInit, Response } from './types'
 import { createFetchError } from './error'
 import { isPayloadMethod, isJSONSerializable, detectResponseType, ResponseType, MappedType } from './utils'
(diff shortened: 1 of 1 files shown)


pi0:
Change makes sense if a generic baseURL is provided and we want to reuse same config to override full URLs but it can also be a breaking change and unintended behavior (adding a URL as path. Imagine BaseURL is like `http://image-cdn.com/s_200/` expecting to append URL to the end...)

danielroe:
It came to my attention because `$fetch` in Nuxt is broken if used with a baseURL because it is created with `.create({ baseURL })`. (So it works with `/`.)

```bash
> require('ufo').withBase('https://test.com', '/')
'https://test.com'
> require('ufo').withBase('https://test.com', '/f')
'/f/https://test.com'
```

This is suggesting to me that maybe it's the behaviour in ufo that we need to fix.

pi0:
Well, the behavior of `ufo.withBase` is basically adding a one-time prefix. Not judging what kind of prefix or suffix we have (with/without protocol) and `/f/http://test.com` is a valid URL too even for Nuxt (A catch-all route can again handle that for something like an analyzer). 

I think this kind of behavior only makes sense if we provided a _default_ baseURL like the case of Nuxt/Nitropack with `.create`. While currently there is no way to distinguish.

danielroe:
I think we could always say that users need to prefix the URL with `/` to get it to be resolved relative to a baseURL.

pi0:
You are right. This also seems URL behavior. Let's do it on ufo then since there is also `/` workaround 👍🏼 https://github.com/unjs/ufo/issues/67

PR #152, opened 2022-10-27
feat: add `status` and `statusText` to fetch errors

danielroe:


src/error.ts
```diff
@@ -21,6 +23,8 @@ export function createFetchError<T = any> (request: FetchRequest, error?: Error,
   Object.defineProperty(fetchError, 'request', { get () { return request } })
   Object.defineProperty(fetchError, 'response', { get () { return response } })
   Object.defineProperty(fetchError, 'data', { get () { return response && response._data } })
+  Object.defineProperty(fetchError, 'status', { get () { return response && response.status } })
+  Object.defineProperty(fetchError, 'statusText', { get () { return response && response.statusText } })
```
danielroe, line 27:
```suggestion
  Object.defineProperty(fetchError, 'status', { get () { return response && response.status } })
  Object.defineProperty(fetchError, 'statusText', { get () { return response && response.statusText } })
  Object.defineProperty(fetchError, 'statusCode', { get () { return response && response.status } })
  Object.defineProperty(fetchError, 'statusMessage', { get () { return response && response.statusText } })
```

src/error.ts
```diff
@@ -5,6 +5,8 @@ export class FetchError<T = any> extends Error {
   request?: FetchRequest
   response?: FetchResponse<T>
   data?: T
+  status?: number
+  statusText?: string
```
danielroe, line 9:
```suggestion
  status?: number
  statusText?: string
  statusCode?: number
  statusMessage?: string
```

pi0 (approved):

PR #199, opened 2022-12-15
fix: improve error message for request errors

silverbackdan:
The error message itself looks erroneous when the fetch fails without any request or response. Usually a network error. This can happen when a user disconnects from the internet or a fetch endpoint has an invalid ssl certfiicate.

Previously the message would be `Failed to Fetch ()` with these empty brackets.

Simple refactor to build the message up in the order that the final string will be and include the brackets only if required.

Thanks for all your work on this project! It's great.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -13,11 +13,11 @@ export class FetchError<T = any> extends Error {
 
 export function createFetchError<T = any> (request: FetchRequest, error?: Error, response?: FetchResponse<T>): FetchError<T> {
   let message = "";
-  if (request && response) {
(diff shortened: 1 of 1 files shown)


silverbackdan:
I've just added 1 more commit now which will still output the request URL if there was no response in brackets. Again, this is useful if the browser intercepted the response, the internet disconnects and such. The error message will still have the useful information about the endpoint that was trying to be fetched.

pi0 (approved):
Thanks!

silverbackdan:
Most welcome @pi0 thanks for your work on OS.

PR #173, opened 2022-11-22
docs: add agent option usage

yuyinws:
add agent option usage on docs. close #161

Issue #161, opened 2022-11-12
HTTP(S) Proxy | Agent With OhMyFetch


BlackYuzia:
Hi there.

Just a short question.
**How we could use `ohmyfetch` with HTTP(S) Proxy | Agent?**

For example.
If we use `axios`, we could use `options.httpsAgent` with [**httpsAgent**](https://www.npmjs.com/package/https-proxy-agent)

In `ohmyfetch`I didn't find any docs about proxy and/or `httpsAgent`

yuyinws:
Need this too.

yuyinws:
It's works for me
```ts
import { Agent } from 'node:http'
import { $fetch } from 'ohmyfetch'

const agent = new Agent({
    host: '127.0.0.1',
    port: 7890,
})

$fetch(url, {
  agent: agent
})
```

yuyinws:
It's works for me on node 16.X

```ts
import HttpsProxyAgent from 'https-proxy-agent'
import { ofetch } from 'ofetch'

await ofetch('user',{
 agent: new HttpsProxyAgent('http://127.0.0.1:7890')
})
```

Rekryt:
Try use changeOrigin
```
devProxy: {
    ...
    '/proxy/example': { target: 'https://example.com', changeOrigin: true }
}
```

BlackYuzia:
Yes, this worked for me too. Thx.

pi0:
PR welcome to the docs.

289997171:
server\api\test\google.get.ts
```javascript
import {HttpsProxyAgent} from "https-proxy-agent";
import {$fetch, ofetch} from "ofetch";
const agent = new HttpsProxyAgent('http://127.0.0.1:7890')

export default defineEventHandler(async (event) => {
    // const res = await ofetch('https://www.google.com', {
    //     agent: agent
    // })
    const res = await $fetch('https://www.google.com', {
        agent: agent
    })
    console.log(res);
    return res
})
```
curl test request
```cmd
curl http://localhost:3000/api/test/google

[...]

PR #202, opened 2023-01-12
docs: add note about http agent

NozomuIkuta:
Related to #161.

This is just a follow-up PR to resolve https://github.com/unjs/ofetch/pull/173#discussion_r1068217168.


NozomuIkuta:
Let me merge this PR now because it doesn't have influence on the actual code.

PR #203, opened 2023-01-12
fix: avoid parsing empty body

### LINKED ISSUE #171: No content (204) responses throw unparseable error
An endpoint that returns a HTTP 204 response without body.

The culprit is this [line of code](https://github.com/unjs/ofetch/blob/main/src/fetch.ts#L144):

```js
   // We override the `.json()` method to parse the body more securely with `destr`
    if (responseType === "json") {
      const data = await context.response.text();
      const parseFunction = context.options.parseResponse || destr;
      context.response._data = parseFunction(data);
    } else if (responseType === "stream") {
      context.response._data = context.response.body;
    } else {
      context.response._data = await context.response[responseType]();
    }
```

Before this, it has to check if the status code is 204 to skip parsing. 

It will also throw the same error on any response without a body.

NozomuIkuta:
Resolves #171.

Still WIP, bug included.

---

In addition, unrelated part of unit test fails even at the latest commit of `main` branch, so this should be fixed first.

```
 FAIL  test/index.test.ts > ofetch > Bypass FormData body
AssertionError: the given combination of arguments (undefined and string) is invalid for this assertion. You can use an array, a map, an object, a set, a string, or a weakset instead of a string
 ❯ test/index.test.ts:134:21
    132|     data.append("foo", "bar");
    133|     const { body } = await $fetch(getURL("post"), { method: "POST", body: data });
    134|     expect(body).to.include("form-data; name=\"foo\"");
       |                     ^
    135|   });
    136|
```

### DIFF
--- a/test/index.test.ts
+++ b/test/index.test.ts
@@ -1,6 +1,6 @@
 import { listen } from "listhen";
 import { getQuery, joinURL } from "ufo";
-import { createApp, eventHandler, readBody, readRawBody, toNodeListener } from "h3";
+import { createApp, eventHandler, readBody, readRawBody, toNodeListener, send, appendResponseHeader } from "h3";
 import { Blob } from "fetch-blob";
 import { FormData } from "formdata-polyfill/esm.min.js";
 import { describe, beforeEach, afterEach, it, expect } from "vitest";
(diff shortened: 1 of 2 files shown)


test/index.test.ts
```diff
@@ -50,6 +75,28 @@ describe("ofetch", () => {
     expect(await $fetch(getURL("binary"))).to.be.instanceOf(Blob);
   });
 
+  it("avoid parsing response when the body is empty", async () => {
+    expect(await $fetch(getURL("1xx"))).not.toThrowError();
```
NozomuIkuta, line 79: An error is thrown at this line.
No-body response can't be fetched...? 🤔 

```
 FAIL  test/index.test.ts > ofetch > avoid parsing response when the body is empty
FetchError: fetch failed ()
 ❯ $fetchRaw2 src/fetch.ts:131:24
    129|     }
    130| 
    131|     context.response = await fetch(context.request, context.options as RequestInit).catch(async (error) => {
       |                        ^
    132|       context.error = error;
    133|       if (context.options.onRequestError) {
 ❯ test/index.test.ts:79:12

Serialized Error: {
  "data": undefined,
  "request": "http://localhost:3000/1xx",
  "response": undefined,
  "status": undefined,
  "statusCode": undefined,
  "statusMessage": undefined,
  "statusText": undefined,
}
```

PR #218, opened 2023-02-28
feat: allow sanitization using ufo

ineshbose:
Motivation:

Using create fetch and composables around `$fetch` in my Nuxt app that have my baseURL, but also not necessarily adding trailing slashes from all points where it is being used which may be required by the server, and rather than providing a single `withTrailingSlash: boolean` option, given all awesome utilities `unjs/ufo` provides, may as well make benefit of all!


ineshbose:
Thank you for the input! I did fear about the bundle size so I was hoping for your thoughts. I'll close this for now. 🙂

Issue #219, opened 2023-02-28
Exported variable 'X' has or is using name 'ResponseMap' from external module


theguriev:
### Environment

Desktop

### Reproduction

I have nothing to provide here

### Describe the bug

Hey guys,

I get an error when I build a project and I seem to know how to fix it (the dirty way). For some reason, if I try to add slightly more complex types than $fetch expects as a generic it all starts to break. 

Screenshots:
<img width="970" alt="image" src="https://user-images.githubusercontent.com/8502021/221943258-bb8c0981-84b3-4e1d-9629-06604b9bff97.png">
<img width="805" alt="image" src="https://user-images.githubusercontent.com/8502021/221943303-078823b6-8885-43eb-8336-ebe25178da3d.png">

The fix is to add the export operator to ResponseMap node_modules/ofetch/dist/error-8a55452d.d.ts
<img width="473" alt="image" src="https://user-images.githubusercontent.com/8502021/221943772-b8a2bbba-9ec4-4378-a7ea-459f8f35e7f9.png">

I'll just add `export` now and publish the package somewhere, but it would be nice to hear your thoughts on how to properly fix it. 

Thank you ❤️

PR #232, opened 2023-04-13
feat: export `ResponseMap` type to allow composition of `ofetch`

codebender828:
This re-exports the `ResponseMap` type. The use case for this is to allow consumers to build a type-safe variant of `ofetch` such we can append methods to a given instance.

Consider this wrapper library below called `api.ts`. In the code below, the `FetchOptions` type from `ofetch` consumes the private type `ResponseMap`. Building declarations for `api.ts` types for this would fail because the `ResponseMap` type is private to `ofetch`. In this case, re-exporting the type fixes the error and declarations can be built in userland.

```ts
// api.ts lib
import { ofetch, FetchOptions } from "ofetch";

export const $fetch = ofetch.create({
  baseURL: SOME_API_BASE_URL,
  headers: {
    Accept: "application/json",
    "Content-Type": "application/json",
  },
  onRequest(context) {
    console.log("context", context.options);
  },
  credentials: "include",
});

/**
 * HTTP API Fetch instance
 * @param request
 * @param options
 * @returns
 */
export function api<T = any, R extends ResponseType = "json">(
  request: FetchRequest,
  options?: FetchOptions<R>
): Promise<MappedType<R, T>> {
  return $fetch<T, R>(request, options) as Promise<MappedType<R, T>>;
}

/* Appends `get` Method to `api` function object */
api.get = function get<T = any, R extends ResponseType = "json">(
  url: string,
  options: Omit<FetchOptions, "method"> = {}
): Promise<MappedType<R, T>> {
  return api(url, {
    ...options,
    method: "GET",
  });
};
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -34,7 +34,7 @@ const textTypes = new Set([
 
 const JSON_RE = /^application\/(?:[\w!#$%&*.^`~-]*\+)?json(;.+)?$/i;
 
-interface ResponseMap {
+export interface ResponseMap {
   blob: Blob;
   text: string;
   arrayBuffer: ArrayBuffer;


danielroe (approved):

madskronborg:
When is this expected to release? :)

We are currently blocked in our migration to Nuxt 3, because our core nuxt module needs to expose a shared api client composable as discussed here: https://github.com/unjs/ofetch/issues/219#issue-1603590283

PR #235, opened 2023-04-27
fix: pass empty object to headers initializer to prevent crash on chrome 49

IlyaSemenov:
ofetch crashes in Chrome 49 (which is the latest available Chrome on Windows XP which is still relevant in certain GovTech) with this message:

```
Failed to construct 'Headers': No matching constructor signature.
```

This is because older Chrome didn't treat `new Headers(undefined)` as `new Headers()`.

This PR fixes this inconvenience with no noticeable burden for modern browsers (or code base readability).

Related: https://github.com/whatwg/fetch/issues/185


danielroe (approved):

pi0 (approved):
Thanks!

PR #239, opened 2023-05-12
docs(readme): use `_data` rather than `data` for raw requests

dargmuesli:
Following https://github.com/unjs/ofetch/pull/49 the README should be updated accordingly.


pi0:
Thanks!

PR #241, opened 2023-05-14
docs: mention `DELETE` is no-retry be default

Lmmmmmm-bb:
In `fetch.ts`, the `DELETE` method `retries` default also is **0**

https://github.com/unjs/ofetch/blob/9a5cd33b568ef70f476bda647cd43b1040aaa479/src/fetch.ts#L99
https://github.com/unjs/ofetch/blob/9a5cd33b568ef70f476bda647cd43b1040aaa479/src/utils.ts#L1-L6

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ await ofetch(...).catch((error) => error.data)
 
 ## ✔️ Auto Retry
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks!

PR #274, opened 2023-08-23
in onRequest method, some one maybe has eq logic like method === 'post'

mythshuazi:
### ❓ Type of change

### 📚 Description

in my project code like this:
```js
onRequest({ options }) {
        // ...
        let { method } = options;

        if (method === "post") {
          options.body.traceId = "xxxxxxxxxxxx"
        }
        // ...
}
```

If the toUpperCase method is executed before
```
context.options.onRequest(context)
```
Then the parameters will never be added to options.body or options.params

### 📝 Checklist

- [ ] I have linked an issue or discussion.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -165,12 +165,12 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
       error: undefined,
     };
 
-    // Uppercase method name
-    context.options.method = context.options.method?.toUpperCase();
-
     if (context.options.onRequest) {
       await context.options.onRequest(context);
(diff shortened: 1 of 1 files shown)


pi0:
Hi dear @mythshuazi thanks for the pull request but actually this behavior is to make sure you write your logic to check method against upper case always and reduce chances of not working code.

mythshuazi:
> Hi dear @mythshuazi thanks for the pull request but actually this behavior is to make sure you write your logic to check method against upper case always and reduce chances of not working code.

@pi0 Thank you for your guidance

Issue #278, opened 2023-08-24
Property 'statusCode' does not exist on type 'FetchError<any>'


misbahansori:
### Environment

Node : v16.14.2
ofetch: "^1.3.2",

### Reproduction

```
try {
    await $fetch("/contact", {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      baseURL: runtimeConfig.public.apiUrl,
      body: form.value,
    });
  } catch (error) {
    if (!(error instanceof FetchError)) throw error;

    if (error.statusCode !== 422) throw error;
              // ^Property 'statusCode' does not exist on type 'FetchError<any>'.ts(2339)
    errors.value = error.data.errors;
                                  // ^Property 'data' does not exist on type 'FetchError<any>'.ts(2339)
  }
  ```

### Describe the bug

After upgrading to version 1.3.2, the following error message appears. I can confirm that the error message does not appear in version 1.3.1.

**Comment by pi0** [comment]
Thanks for making issue. You can use new `IFetchError` type to fix. I will try to find a way to fix `FetchError` (class) types. We had to make this change to avoid unnecessary runtime overhead...

**Comment by sharifzadesina** [comment]
It was actually working before.
Now, even I can't access `response` property. Actually non of `IFetchError` properties.

**Comment by johannschopplich** [comment]
@misbahansori @sharifzadesina I had the same problem and provided a PR to ensure TypeScript understands and provides properties like `statusCode`, `statusMessage` etc.: #279 

For my own purposes before the PR gets merged (or a better solution found 😉), you can re-export the type:
```ts
export type { IFetchError as FetchError } from 'ofetch'
```

Which I did in my Nuxt API Party plugin: https://github.com/johannschopplich/nuxt-api-party/commit/181a5a0dfb1e494c571bd84f09cb9a5b9d61fd86

misbahansori:
In the meantime, does anyone have a good approach to narrow the types within the catch block? Since the error type is `unknown` we need to narrow it down. currently, I'm using `error instanceof FetchError` which has a runtime implication.

pi0:
Should be resolved by https://github.com/unjs/ofetch/pull/279

PR #279, opened 2023-08-24
fix: augment `FetchError` type to include `IFetchError`

johannschopplich:
### 🔗 Linked issue

#278

### ❓ Type of change

### 📚 Description

With the latest ofetch version, the `FetchError` does implement the `IFetchError` interface, but TS won't provide the types. Thus, only `message`, `stack` etc. properties are available.

Since the `FetchError` will be initiated by `createFetchError`, we can add the properties from the interface directly in the class definition, without the need of setting them in the constructor. With these changes, TypeScript understands.

### 📝 Checklist

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -17,15 +17,6 @@ export interface IFetchError<T = any> extends Error {
 }
 
 export class FetchError<T = any> extends Error implements IFetchError<T> {
-  request?: FetchRequest;
-  options?: FetchOptions;
-  response?: FetchResponse<T>;
-  data?: T;
-  status?: number;
-  statusText?: string;
-  statusCode?: number;
-  statusMessage?: string;
-
   constructor(message: string, opts?: { cause: unknown }) {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for this PR dear @johannschopplich but splitting the types out of the class was intentional. (sadly) esbuild adds them as public accessors which makes them (wrongly) visible on stack traces while they should be getters only..

johannschopplich:
Oh, I see! I wasn't aware of that issue. Maybe there's another way of merging the types then. 🤔  Feel free to close this PR then.

johannschopplich:
Edit: I have updated this PR to use a type-only approach: We can augment `FetchError` type to include all `IFetchError` properties.

```ts
// Augment `FetchError` type to include `IFetchError` properties
export interface FetchError<T = any> extends IFetchError<T> {}
```

pi0 (approved):
💯

PR #309, opened 2023-10-27
refactor: use `AbortSignal.timeout` instead of `AbortController` and `setTimeout`

zhings:
AbortSignal.timeout() is more simple, and no need to clearTimeout().


pi0:
Thanks! Checking if we need any more polyfills for legacy node.

PR #343, opened 2023-12-12
feat: cap maximum number of `retries` at `25`

johannschopplich:
### 🔗 Linked issue

Related:
- https://github.com/johannschopplich/nuxt-api-party/security/advisories/GHSA-q6hx-3m4p-749h

### ❓ Type of change

### 📚 Description

This PR adds a sanity default value for the configurable `retries` option. 

This enhancement is suggested by @pi0 following the security issue detected by @OhB00. In the Nuxt module API Party the fetch options were passed to a server route, which consumed the fetch options as-is. This enabled attackers to force a `retry` value of whatever.

It might be best to set a maximum number of retries. Not only for security reasons, but also for developers as a safety measures.

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ await ofetch("http://google.com/404", {
 });
 ```
 
-> [!TIP]
+> [!NOTE]
 > To mitigate the risk of overloading resources, the number of retries is capped at 25.
 
 ## ✔️ Timeout


README.md
```diff
@@ -136,6 +136,9 @@ await ofetch("http://google.com/404", {
 });
 ```
 
+> [!NOTE]
```
pi0, line 139: I think we can use a normal paragraph since no user action is involved...

src/fetch.ts
```diff
@@ -46,11 +49,15 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
         context.error.name === "AbortError" &&
         !context.options.timeout) ||
       false;
+
     // Retry
     if (context.options.retry !== false && !isAbort) {
       let retries;
       if (typeof context.options.retry === "number") {
         retries = context.options.retry;
+        if (retries > MAX_RETRIES) {
```
pi0, line 58: What about introducing new `maxRetries` option to allow setting the default?

README.md
johannschopplich: Good point!

src/fetch.ts
johannschopplich: I wonder if this defeats the purpose of this PR. My understanding was to include a sensible default for maximum retries. Instead of a `maxRetries` option, `retries` basically does the same, right?
If you dislike the idea of an internal maximum, which I completely get, please close this PR. 🙂

Issue #319, opened 2023-11-11
Allow having interceptos from both defaults and per request


enkot:
### Describe the feature

I need to log some data on each request/response, transform the request url (e.g. `/pets/{petId}` -> `/pets/1`) etc. 
If I add `onRequest` to the `$fetch.create` and then add `onRequest` to the actual client:
```ts
const $myFetch = $fetch.create({
  onRequest() {
     // Log data
  }
})

$myFetch({
  onRequest() {} // logging is lost
})
```
handler from `$fetch.create` will be replaced by the handler from `$myFetch`. 

To get around this we need to create the wrapper around $fetch:
```ts
function $myFetch(url: string, options: FetchOptions) {
  return $fetch(url, {
    ...options,
    onRequest(ctx) {
      // Log data and transform url
      // ...
      return options.onRequest?.(ctx)
    }
  })
}
```

What do you think about adding some sort of interceptors "merging strategy" option?

It can be done based on the idea of this PR https://github.com/unjs/ofetch/pull/353.
If we treat interceptors as an array, we can allow merging arrays from `$fetch` and `$fetch.create`. So that interceptors from `$fetch` and `$fetch.create` can be enforced to append or prepend to the resulting array of interceptors (similar to Nuxt plugins):

```ts
const $myFetch = $fetch.create({
  onRequest: [
    (ctx) => { /* Handler 1 */ }, // same as "enforce: 'default'"
    {
      enforce: 'post',
      handler: (ctx) => { /* Handler 2 */ }
    }
  ]
})

$myFetch({
  onRequest: [
    // Will be appended 
    (ctx) => { /* Handler 3 */ },
    // If you need to prepend in some scenarios
    {
      enforce: 'pre',
      handler: (ctx) => { /* Handler 4 */ }
    }
  ]
})

// Interceptors will be executed in this order:
/*
Handler 4
Handler 1
Handler 3
Handler 2
*/
```

#### Alternative Original idea
```ts
const $myFetch = $fetch.create({
  onRequest: {
    strategy: 'before', // 'overwrite' | 'before' | 'after' | 'manual', 'overwrite' by default
    handler(ctx) { ... }
  }
})

$myFetch({
  onRequest() {...}, 
})
```
so that interceptors from `$myFetch` are called after interceptors from `$fetch.create`. 

### Additional information

- [X] Would you be willing to help implement this feature?

enkot:
[...]

PR #353, opened 2024-01-25
feat: support interceptors arrays

antoinerey:
### 🔗 Linked issue

None.

### ❓ Type of change

### 📚 Description

This commit allows passing interceptors as single functions (current behaviour), or arrays of functions (new behaviour). This makes it much easier to execute multiple functions on the consumer side. The alternative would be to compose all functions into one, but that's more complicated in my opinion.

With this change, the following becomes possible.

```ts
$fetch('/api/endpoint', {
  onRequest: [
    () => { /* Do something */ },
    () => { /* Do something else */ },
  ],
  onResponse: [
    () => { /* Do something */ },
    () => { /* Do something else */ },
  ],
})
```

Note that calling `$fetch.create()` and then passing interceptors when calling `$fetch()` will keep overriding the interceptors as the current behaviour. I did not want to go too much into there as I'm not sure what would be the expected behaviour. Plus, continuing to override the interceptors means that this PR is not a breaking change.

### 📝 Checklist

- [ ] I have linked an issue or discussion.


src/types.ts
```diff
@@ -130,3 +134,5 @@ export type FetchRequest = RequestInfo;
 export interface SearchParameters {
   [key: string]: any;
 }
+
+type Arrayable<T> = T | Array<T>;
```
antoinerey, line 138: I'm not sure about this one. It seems simpler as interceptor types are only defined once, but the more verbose approach might also be viable.

```ts
onRequest?: 
  | ((context: FetchContext) => Promise<void> | void)
  | Array<(context: FetchContext) => Promise<void> | void>;
```

antoinerey:
Friendly ping. 🤗 Is there anything I can do to make this PR move forward?

enkot:
@antoinerey @pi0 
I think it might be a better alternative to my proposal (https://github.com/unjs/ofetch/issues/319) if we treat interceptors as an array and allow merging arrays from `$fetch` and `$fetch.create`. Then we can append or prepend interceptors from `$fetch` to interceptors from `$fetch.create`:
```ts
  // Set default strategy and handlers
  onRequest: {
    strategy: 'append', // 'overwrite' | 'prepend' | 'append'
    handlers: [
      (ctx) => { /* Do something */ }
    ]
  }
})

// Will be appended 
$myFetch({
  onRequest() { /* Do something */ }
})

$myFetch({
  onRequest: {
    // override default strategy
    strategy: 'prepend',
    handlers: [
      (ctx) => { /* Do something */ }
    ]
  }
})
```
Now I don't like my PR anymore 😅

Issue #160, opened 2022-11-12
onResponseError


maksimdrosdov:
hello, on submit, if the tokens are valid, everything works correctly and Swal displays a message, if the token is not valid, a refresh fires, the form submit fires again, but Swal no longer displays a message. How to deal with it?

```
 async onResponseError(ctx: FetchContext): Promise<any> {
    if (ctx.response.status === 401) {
      const { error } = await this.refreshToken();

      if (!error) {
        // eslint-disable-next-line no-return-await
        return await this.fetcher(ctx.request, ctx.options);
      }
    }
  }
```
repository

```
async addProductToCart(body: any, filter: any = {}): Promise<any> {
    return this.fetcher(this.urlGenerator.addProductToCart(), {
      method: 'POST',
      body,
      params: filter
    });
  }
```

my function

```
const onSubmit = handleSubmit(async () => {
    const {
      error,
      pending
    } = await useLazyAsyncData(
      'updateUserProfile',
      () => httpRepository.addProductToCart(form),
      {
        initialCache: false
      }
    );

    isLoading.value = pending.value;

    if (!error.value) {
      Swal.fire({
        toast: true,
        position: 'top-right',
        text: 'Данные успешно обновлены!',
        icon: 'success',
        timer: 3000,
        showConfirmButton: false
      });
    }
  });
```

NozomuIkuta:
I feel this issue is potentially related to and/or duplicated of #114.
But, we appreciate if you provide a minimal reproduction by which everyone can see what you see.

linshujuan:
@maksimdrosdov Hi, do you solve this issue, I also meeting this problem right now

TusharRoy23:
Any update on that?

selivanovvv:
hey, did u solve the issue?

Atinux:
I believe this is related to the fact that we don't look at what `onResponseError` can return: https://github.com/unjs/ofetch/blob/411904d3bf8c5c2f8009adf1e24af2a9fd7f249d/src/fetch.ts#L283

This means that we ignore what you can return.

I think we should change the behavior in order to transform the response based on it, similar to Axios.

trandaison:
I believe this issue is same as #224.

Have you tried to modify the `context.response` in `onResponseError` yet? It should work.

https://github.com/unjs/ofetch/issues/224#issuecomment-1592183119

cgnz0r:
[...]

PR #357, opened 2024-02-02
feat: have both local and global interceptors

enkot:
### 🔗 Linked issue

Closes
- https://github.com/unjs/ofetch/issues/319
- https://github.com/unjs/ofetch/issues/160
### ❓ Type of change

### 📚 Description

Currently interceptors specified in $fetch() override the interceptors in $fetch.create. 
This PR implements:
1. The interceptor "merge strategy" options that allow you to have both default and per-request interceptors. You can add interceptor as an object with `handler` and `strategy` fields when creating new client using `$fetch.create`. Available strategies: 
- 'overwrite' - current behaviour
- 'after' - interceptor in $fetch will be called after the interceptor in $fetch.create
- 'before' - interceptor in $fetch will be called before the interceptor in $fetch.create
- 'manual' - interceptor in $fetch.create decides how/when/if call interceptor from $fetch

In all cases interceptor has second parameter which can be:
- response data
- data return by the previous interceptor 

_Names are subject to change_ 👀 

2. It is now allowed to return new response data from the `onResponse`, `onResponseError` interceptors.
If the function doesn't return anything or returns `undefined` - the response data will not be overwritten.

Example:
```ts
const myFetch = $fetch.create({
  onResponse: {
    strategy: "after",
    handler(ctx, data) {
      return 'global: ' + data
    },
  }
})

const r = await myFetch<string>("http://httpstat.us/200", {
  onResponse(ctx, data) {
    return 'local -> ' + data
  }
})

console.log(r) // will print "local -> global: 200 OK"
```

The open question is how to type the "data" parameter, currently it has type `any`.

We can also introduce the `defaultStrategy` option.

### 📝 Checklist

### DIFF
--- a/playground/index.ts
+++ b/playground/index.ts
@@ -1,8 +1,21 @@
 import { $fetch } from "../src/node";
 
+const myFetch = $fetch.create({
+  onResponse: {
+    strategy: "after",
+    handler(ctx, data) {
+      return 'global: ' + data
+    },
+  }
+})
+
 async function main() {
+  const r = await myFetch<string>("http://httpstat.us/200", {
+    onResponse(ctx, data) {
+      return 'local -> ' + data
+    }
+  });
   // const r = await $fetch<string>('http://google.com/404')
-  const r = await $fetch<string>("http://httpstat.us/500");
   // const r = await $fetch<string>('http://httpstat/500')
   // eslint-disable-next-line no-console
   console.log(r);
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -7,6 +7,7 @@ import {
   isJSONSerializable,
   detectResponseType,
   mergeFetchOptions,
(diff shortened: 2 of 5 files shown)


pi0:
Thank you for PR dear @enkot while some changes are nice, i found it little over complex and some against the design patterns of ofetch. 

For the initial step, I went with https://github.com/unjs/ofetch/pull/353 which allows an array of interceptor options.

A quick explanation of other parts:

- ofetch does not allow any kind of global behavior change  (this is by-design). we might (rightfully) need global hooks that can log requests but they shouldn't be able to block or change global fetch behavior so they are not exactly hooks. 
- with the array format, the order of hooks is already explicit
- we can support returning custom Responses by all hooks, as a followup of #353 in callHook internal util.

Issue #358, opened 2024-02-05
feat: add `retryCb` option


enkot:
### Describe the feature

There are situations when we need to retry the request based on the special code returned by the API. 
I suggest adding a `retryCb` (or `retryIf` or something) option, which is a function that takes an error object and returns a boolean value of true if the request needs to be retried:

```ts
await $fetch('/cart', {
  retryCb(err) {
    return err.errorId === 'SPECIAL_CODE'
  }
})
```

### Additional information

- [X] Would you be willing to help implement this feature?

Issue #355, opened 2024-01-27
Ability to retry a request after an error


CernyMatej:
### Describe the feature

I believe it would be beneficial to have the ability to easily retry a failed request based on custom logic defined in an interceptor. For instance, in the case of a `401` response, you could refresh your session in an interceptor and then, if the refresh is successful, you could retry the original request easily by calling `retry()`.

What do you think?

### Additional information

- [X] Would you be willing to help implement this feature?

devmini1203:
I very much agree with you.
In axios, it is easy for requests that failed in 401 error
I hope this function will be applied to ohmyfetch and can be used conveniently in nuxt3

huxc:
Is this feature currently available? I tried it and it didn't work

huxc:
Is this feature currently available?

zawasp:
This would also help with Cloudflare Pre-clearance:
https://blog.cloudflare.com/integrating-turnstile-with-the-cloudflare-waf-to-challenge-fetch-requests/

zay0od:
Is this feature currently available? or still?

devmini1203:
> 이 기능을 현재 사용할 수 있나요? 아니면 아직도 사용할 수 있나요?

I don't think I can use it yet..

PR #359, opened 2024-02-05
feat: retry callback

enkot:
### 🔗 Linked issue

https://github.com/unjs/ofetch/issues/358

### ❓ Type of change

### 📚 Description

This PR adds new `retryCb` option which is a function that takes a fetch context object and returns a boolean, `true` if the request needs to be retried:

```ts
await $fetch('/cart', {
  retryCb(ctx) {
    return ctx.error.errorId === 'SPECIAL_CODE'
  }
})
```

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,13 @@ The default for `retry` is `1` retry, except for `POST`, `PUT`, `PATCH`, and `DE
 
 The default for `retryDelay` is `0` ms.
 
+You can also use `retryCb` function to retry on some condition. It takes a fetch context object as an argument.
+
 ```ts
 await ofetch("http://google.com/404", {
   retry: 3,
   retryDelay: 500, // ms
+  retryCb: (ctx) => ctx.error.code === '007'
 });
 ```
 
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -56,12 +56,14 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
       }
 
       const responseCode = (context.response && context.response.status) || 500;
-      if (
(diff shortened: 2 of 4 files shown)


gustavopch:
My 2 cents: `shouldRetry` seems to better describe what this does.

enkot:
@pi0 Could you please check? I think it's a really flexible solution.

CernyMatej:
I love this idea. Probably link my issue as well as this should cover that use case too: https://github.com/unjs/ofetch/issues/355

AryanShrivastava18:
@enkot By when can we expect this change to be the part of the release? Is there anything I can give a support for to push it to release?

lacherogwu:
retryCount will be essential 🙏

fredericDelaporte:
@gustavopch,
> My 2 cents: shouldRetry seems to better describe what this does.

The PR does not add a new option. It alters an existing one which currently can only be `number | false`, for it to also accept a callback returning a boolean or promise of a boolean. See [here](https://github.com/unjs/ofetch?tab=readme-ov-file#%EF%B8%8F-auto-retry) if you are not aware of that existing option. (Maybe @lacherogwu too.)

@enkot, thanks for the PR, I too would need this feature. I hope it can get merged. It would likely help if you could resolve the conflicts it has with current code.

But maybe it will not entirely cover my use case: an API that may return 401 on Bearer authentication with an access token, and having an endpoint taking refresh tokens for renewing the authentication. With your callback, I could attempt to renew the token
and retry the query if the refresh succeeded, assuming there is a way to alter the headers to be sent for the retry, in order to use the new bearer token.

PR #371, opened 2024-03-05
feat(types): typed api definition

### LINKED ISSUE #364: Moving $Fetch (with type safety inference) from nitro to ofetch
### Describe the feature

Hi! I have been using nuxt and ofetch for a while. The ofetch package(independently) lacks the types safety I get when working with ofetch with nitro. I have to specify types for response on every request manually. So, I looked up in the repo and found out that types inference based on schema was a part of nitro not ofetch. I think those types should belong here in ofetch repo so, if we work independently with ofetch we can get type safety.

I open a dicussion in Nitro repo, here https://github.com/unjs/nitro/discussions/2157

I have been working on it for past couple of days. It is almost done. There are few thing to fix and discuss.

While I was implementing it. I thought it is also time to change the structure of `InternalApi` so that it will now support more types like body/query/params.

Here is now schema I come up with.

```ts
interface ApiDefinition {
  "/api/v1": {
    default: {
      response: { message: string };
    };
  };
  "/api/v1/auth/register": {
    post: {
      response: { message: string };
      request: {
        body: {
          name: string;
          email: string;
          username: string;
          password: string;
        };
      };
    };
  };
  "/api/v1/users/search": {
    get: {
      response: { users: { username: string }[] };
      request: {
        query: {
          username: string;
        };
      };
    };
  };
  "/api/users/:username": {
    get: {
      response: { user: { username: string; name: string; isFriend: boolean } };
      request: {
        params: {
          username: string;
        };
      };
    };
    post: {
      response: { message: string };
      request: {
        params: {
          username: string;
        };
      };
    };
  };
}
```

You can check out my version of implementation in 
https://github.com/anuragkumar19/ofetch

There are some examples embedded in the `types.ts` for testing while development which will be removed in final version.

I have also examples in `playground/index.ts` for convention. 

I think it will be totally backward compatible but we will discuss. Also compatible with nitro because it fully override `$Fetch`, also after this change we can fully drop `types/fetch.ts` in nitro. 

This will solve many issues opened in nitro and nuxt repo for type-safety with request object.

I will open a PR soon. I need to clean few this up first.

### Additional information

anuragkumar19:
### 🔗 Linked issue

#364 

### ❓ Type of change

### 📚 Description

This change will allow developers to define their api schema at one place and get type-safe fetch client.

Resolved #364 

### 📝 Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,27 +2,81 @@
 // $fetch API
 // --------------------------
 
-export interface $Fetch {
-  <T = any, R extends ResponseType = "json">(
-    request: FetchRequest,
-    options?: FetchOptions<R>
-  ): Promise<MappedResponseType<R, T>>;
-  raw<T = any, R extends ResponseType = "json">(
-    request: FetchRequest,
-    options?: FetchOptions<R>
-  ): Promise<FetchResponse<MappedResponseType<R, T>>>;
+// TODO: set default to any for backward compatibility
+export interface $Fetch<DefaultT = unknown, A extends object = InternalApi> {
+  <
+    T = DefaultT,
+    ResT extends ResponseType = "json",
+    R extends ExtendedFetchRequest<A> = ExtendedFetchRequest<A>,
+    M extends
+      | ExtractedRouteMethod<A, R>
+      | Uppercase<ExtractedRouteMethod<A, R>> =
+      | ExtractedRouteMethod<A, R>
+      | Uppercase<ExtractedRouteMethod<A, R>>,
+  >(
+    request: R,
+    opts?: FetchOptions<
+      ResT,
+      {
+        method: M;
+        query: TypedInternalQuery<R, A, Record<string, any>, Lowercase<M>>;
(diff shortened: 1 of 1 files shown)


danielroe (commented):
This looks really great, and thank you so much for your work on this.

I think it makes sense for these types to live in `ofetch`, and I think the schema makes sense:

```ts
[url: string]: {
    [key: 'default' | HTTPMethod]: {
      request: {
        body?: BodyType;
        query?: QueryType
      };
      response: ResponseType;
    };
  };
```

It might also allow us to do things like define error types in future.

My one note is that I think we should spend some additional time thinking about the `params` implementation as this will require runtime code changes as well as just types.

Finally, having said that, the tricky thing will be to properly handle type inference when (for example) hitting and endpoint that has multiple methods with multiple valid request types, each mapping to a method.

Writing those types will be complex.

For that reason, I would like to see some type tests bundled with this PR, for example similar to ones we have in Nitro, covering the way these types are used and ensuring we don't cause regressions.

cain:
really looking forward to use this!

PR #376, opened 2024-03-18
feat: have both local and global interceptors

enkot:
### 🔗 Linked issue

https://github.com/unjs/ofetch/issues/319

### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Currently interceptors specified in $fetch() override the interceptors in $fetch.create.

This PR adds support for interceptors as an array, so we can merge arrays from $fetch and $fetch.create. 
It also allows set an `enforce` option, so that interceptors from $fetch and $fetch.create can be enforced to append or prepend to the resulting array of interceptors (similar to Nuxt plugins):
```ts
  onRequest: [
    (ctx) => { /* Handler 1 */ }, // same as "enforce: 'default'"
    {
      enforce: 'post',
      handler: (ctx) => { /* Handler 2 */ }
    }
  ]
})

$myFetch({
  onRequest: [
    // Will be appended 
    (ctx) => { /* Handler 3 */ },
    {
      enforce: 'pre',
      handler: (ctx) => { /* Handler 4 */ }
    }
  ]
})

// Interceptors will be executed in this order:
/*
Handler 4
Handler 1
Handler 3
Handler 2
*/
```

### Note
This is implemented as a braking change, but we can add the `overrideInterceptors` option to not break the current behavior.

### 📝 Checklist

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -162,7 +162,6 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
       );
     } catch (error) {
       context.error = error as Error;
-      console.log("context.onRequestError", context.options.onRequestError);
       await callInterceptors(context, "onRequestError");
       return await onError(context);
     }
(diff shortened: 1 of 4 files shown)


pi0:
Sorry, this got stalled, we can do it based on new `callHooks` util. Let's track via #319, if had time to do it it is more than welcome ❤️

Issue #295, opened 2023-09-11
Usage with MSW / patched fetch


meesvandongen:
### Environment

Node

### Reproduction

https://codesandbox.io/p/sandbox/stoic-https-6gjdtz?file=/index.js:14,16

```ts
import { setupServer } from "msw/node";
import { http, HttpResponse } from "msw";
import wretch from 'wretch'
import { ofetch, fetch, createFetch } from 'ofetch'
import got from 'got'

const server = setupServer(
    http.get('https://example.com', () => {
        return HttpResponse.text('1')
    })
)
server.listen();

console.table([
 ['fetch', await globalThis.fetch('https://example.com').then(res => res.text()).then(res => res === '1')],
 ['ofetch fetch', await fetch('https://example.com').then(res => res.text()).then(res => res === '1')],
 ['ofetch createFetch', await createFetch()('https://example.com').then(res => res === '1')],
 ['ofetch', await ofetch('https://example.com').then(res => res === '1')],
 ['wretch', await wretch('https://example.com').get().text().then(res => res === '1')],
 ['got', await got.get('https://example.com').text().then(res => res === '1')],
])
```

![CleanShot 2023-12-23 at 14 00 06@2x](https://github.com/unjs/ofetch/assets/904724/f2f9c044-59c3-4291-a19e-f44df7d93aa0)

### Describe the bug

When other tools patch fetch, the global instance of ofetch will have the original `fetch` instance. This will cause tools such as msw to not be able to intercept requests with ofetch. A workaround could be to only get the globalThis.fetch at request time if no other fetch is specified.

### Additional context

Users can work around this at this time by using the `createFetch` helper function.

mohammadGh:
Thanks @meesvandongen for opening this issue, which is also a problem we have encountered. Dear @pi0 please let us know what is the best and efficient temporary solution to solve this problem until it is resolved?

edumudu:
For anyone coming here, I found a hacky and pretty bad but functional way to workaround this for now using the `Proxy` constructor

```ts
function createMyFetch() {
  const newFetch = createFetch({
    fetch: globalThis.fetch,
    Headers: globalThis.Headers,
  });

  return newFetch.create({
    // ...default options
  });
}

const original = createMyFetch();

export const http = new Proxy(original, {
  apply(_, thisArg, argumentsList) {
    return Reflect.apply(createMyFetch(), thisArg, argumentsList);
  },
});
```

[...]

PR #377, opened 2024-03-21
fix: use wrapper to allow patching global `fetch`

dwightjack:
### 🔗 Linked issue

https://github.com/unjs/ofetch/issues/295

### ❓ Type of change

### 📚 Description

I was investigating the usage of MSW on Nuxt 3 and encountered #295. Looking at the source code, I think that a possible solution is to define the local `fetch` constant as _getter_ instead of a reference to `globalThils.fetch`. In this way, 3rd party library can patch and overwrite the global object without us referencing an outdated function:

```ts
// Before
export const fetch = _globalThis.fetch

// After
export const fetch = (...args: Parameters<typeof globalThis.fetch>) => _globalThis.fetch(...args)
```

At the moment, I cannot think of possible breaking changes or side effects caused by this change, but I might miss something.

### 📝 Checklist


rinux55:
https://github.com/nuxt/test-utils/issues/775 just mentioning this one here, thanks for taking this up!

pi0:
Thanks for PR. I was thinking of same solution too. I agree it is needed but (sadly) means 1) we allow patching behavior of a standard primitive 2) adds small call overhead.

@dwightjack Have you tried this solution btw? I guess we might need to patch the node entrypoint as well.

dwightjack:
> Thanks for PR. I was thinking of same solution too. I agree it is needed but (sadly) means 1) we allow patching behavior of a standard primitive 2) adds small call overhead.

1. Not a fan as well, but I guess it's the only way for them to intercept fetch calls
2. I am not sure how much this could impact the performances 🤔 For sure, it's an unneeded change for many use cases.

> @dwightjack Have you tried this solution btw? I guess we might need to patch the node entrypoint as well. (also in `createFetch` we persist the `fetch` variable)

Oh, I missed that part. I updated the PR (https://github.com/unjs/ofetch/pull/377/commits/4026829a54b6d08275c01cb29524ecd7da708359). 

I wasn't able to make a full reproduction of Nuxt + MSW on StackBlitz (I guess there are some problems with service workers), so I pushed it to a GitHub repository: https://github.com/dwightjack/ofetch-msw-reproduction

Some notes:
1. The repository includes the packed version of ofetch built from this PR (I am not sure if there's a more _elegant_ way to do this) and I defined an override for every instance of ofetch. 
1. I implemented MSW in a Nuxt plugin (server + client) and a Nitro plugin (just in case)
4. Mocking an API call with `useFetch` seems to work for client-only calls, but fails for server calls
5. In the `node-app` folder I ported the small node application linked in the original issue using both the un-modified and the patched versions of ofetch. In that case, the patched version seems to work correctly, so I am not sure why it is not working on Nuxt. 🤔 Could it be related to `also in `createFetch` we persist the `fetch` variable`?

vanstinator:
I'm working to add support to `otel-cf-workers` and auto-instrumentation of `fetch` has no effect there for folks using `ofetch`, because by the time `globalThis.fetch` is patched by `otel-cf-workers` `ofetch` has already initialized and doesn't pick up on the new instrumented `fetch` object assigned to `globalThis.fetch`.

I think the work being done here would unblock this use-case too!

MuhammadM1998:
Related https://github.com/nuxt/nuxt/discussions/24519

pi0 (approved):
Thanks ❤️

PR #392, opened 2024-05-04
fix: normalize headers object before onRequest (nuxt/nuxt#27042)

cmgdragon:
### ❓ Type of change

### 📚 Description

Resolves nuxt/nuxt#27042

It would be convenient to normalize the headers object before running the onRequest interceptor so there's no conflicts  between the retried requests

### 📝 Checklist

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -62,6 +62,11 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
           ? context.options.retryStatusCodes.includes(responseCode)
           : retryStatusCodes.has(responseCode))
       ) {
+        // Normalize headers
+        if (context.options.headers instanceof Headers) {
+          context.options.headers = Object.fromEntries(context.options.headers);
+        }
+
         const retryDelay = context.options.retryDelay || 0;
         if (retryDelay > 0) {
           await new Promise((resolve) => setTimeout(resolve, retryDelay));


manniL:
Looks reasonable to me! 👍🏻

cmgdragon:
Heads up! @manniL , I just realized, this approach will only work either using the Headers constructor or setting the properties in lowercase. If we set a header in uppercase inside the `onRequest`, both properties will remain in the headers object and will be merged together after the retry.

For ensuring the headers object always keeps the same format and it isn't case-sensitive, we could add some code after and before the intercerpor... Something like shown below 👇

`utils.ts`
```
export function normalizeHeaders(headers: any) {
  if (headers instanceof Headers) {
    headers = Object.fromEntries(headers)
  }
}
```
`fetch.ts`
```
const headersCopy = context.options.headers ? 
  JSON.parse(JSON.stringify(context.options.headers)) : {};

if (context.options.onRequest) {
  await context.options.onRequest(context);
}

// Normalize & merge headers
normalizeHeaders(context.options.headers);
if (context.options.headers) {
  context.options.headers = {
    ...headersCopy,
    ...Object.fromEntries(
      Object.entries(context.options.headers)
        .map(([k, v]) => [k.toLocaleLowerCase(), v])
    )
  };
}

```

But I don't know if we'd complicating the code too much with that, what do you think?

pi0:
Hi dear @cmgdragon i appreciate the attempt to fix but please first help to make a minimal reproduction of issue using `ofetch` alone and submit an issue report in this repo 🙏🏼 (we might think to solve this differently after discussion)

PR #398, opened 2024-05-24
fix: use  `response._bodyInit` as fallback for react-native and qq

### LINKED ISSUE #324: `undefined` responses when used with React Native
### Environment

React Native (all versions)

### Reproduction

It is difficult to provide a react-native starter as I'm not a specialist, and it will require that you install and setup everything to start a react-native project. Instead, you could rely on my detailed investigation

### Describe the bug

Hi,

in the light of [this issue](https://github.com/unjs/ofetch/issues/293),

Looking deeper into the commit linked by @pi0 to close the issue, and having the same issue as @gtjamesa (responses are all `undefined`), it seems that this commit will not fix the `undefined` responses issue.

I dug into `ofetch` source code and did my own investigation. It appears that there is a boolean `hasBody` which checks if `response.body` exists, [as expected in fetch Response spec](https://developer.mozilla.org/en-US/docs/Web/API/Response), see: https://github.com/unjs/ofetch/blob/bb2d72baa5d3f332a2185c20fc04e35d2c3e258d/src/fetch.ts#L172-L173

Yet, only these fields are available in the `response` object returned by `fetch` in React Native:
`_bodyBlob`, `_bodyInit`, `bodyUsed`, `headers`, `status`, `statusText`, `type`.

Indeed, react-native [uses](https://github.com/facebook/react-native/commit/bccc92dfdd2d85933f2a9cb5c8d1773affb7acba) [whatwg-fetch polyfill](https://github.com/JakeChampion/fetch).

We can easily see [in whatwg-fetch source code](https://github.com/JakeChampion/fetch/blob/main/fetch.js) that all these keys are defined by the polyfill on `this`, but **not** `body`.

The correct answer would be to create an issue on `whatwg-fetch` so that they update their code and comply with Fetch API, and then create an issue on React Native to use this updated library, and then update React Native. Nevertheless, this process can take a while, and (as it is the case for our projects) many React Native projects are stuck to older versions because they depend on other libs that breaks and haven't updated yet and an update to the latest is very difficult.

The shorter answer would be to adapt `ofetch` so that it also checks specific `whatwg-fetch` fields in the response to know if a body is available.

What do you think ?

### Additional context

_No response_

### Logs

_No response_

joshmossas:
### 🔗 Linked issue

Fixes #324 

### ❓ Type of change

### 📚 Description

This pr is essentially a one line change to add react native support. It checks `response._bodyInit` in addition to `response.body` to see if a body exists in the response.

Normally I would say that this is an upstream issue that react native should fix, but the reality is fetch has not been spec compliant in react native for a long time, and it doesn't seem like there has been any meaningful progress in that department either. So I don't think it's practical to wait on them. 

I would also argue that even though this is a runtime specific check it shouldn't add any overhead for maintainers because its such a small change. I've also included comments explaining what the additional `_bodyInit` check is for and under what conditions it can be removed.

Additionally this repo does have a precedent for runtime specific check. In `./src/fetch.ts` line 84, there is a check for a chromium specific API. So this change would be in line with that.
https://github.com/unjs/ofetch/blob/f7f946395544f3ab39fd2e5c61110b5371d7e9dd/src/fetch.ts#L84

In all I think this is a worthwhile change as it would let users use the same fetch library across, node, browsers, and react native.

### 📝 Checklist

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -196,6 +196,7 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
     const hasBody =
       (context.response.body ||
         // https://github.com/JakeChampion/fetch/issues/1454
+        // https://github.com/unjs/ofetch/issues/294
         (context.response as any)._bodyInit) &&
       !nullBodyResponses.has(context.response.status) &&
       context.options.method !== "HEAD";
@@ -218,6 +219,7 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
           context.response._data =
             context.response.body ||
             // https://github.com/JakeChampion/fetch/issues/1454
+            // https://github.com/unjs/ofetch/issues/294
             (context.response as any)._bodyInit;
           break;
         }
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -185,7 +185,11 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
     }
 
     const hasBody =
-      context.response.body &&
+      (context.response.body ||
+        // React Native whatwg-fetch check. Can be removed when the following occurs
+        // 1) https://github.com/JakeChampion/fetch/issues/1454 is fixed
+        // 2) React Native upgrades to the whatwg-fetch version that has this fix
+        (context.response as any)._bodyInit) &&
       !nullBodyResponses.has(context.response.status) &&
       context.options.method !== "HEAD";
     if (hasBody) {
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -218,8 +218,7 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
         }
         case "stream": {
           context.response._data =
-            context.response.body ||            
-            (context.response as any)._bodyInit; // (see refs above)
+            context.response.body || (context.response as any)._bodyInit; // (see refs above)
           break;
         }
         default: {
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -195,8 +195,9 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
 
     const hasBody =
       (context.response.body ||
-        // https://github.com/JakeChampion/fetch/issues/1454
+        // https://github.com/unjs/ofetch/issues/324
         // https://github.com/unjs/ofetch/issues/294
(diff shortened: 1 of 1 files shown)


src/fetch.ts
```diff
@@ -194,7 +194,9 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
     }
 
     const hasBody =
-      context.response.body &&
+      (context.response.body ||
+        // https://github.com/JakeChampion/fetch/issues/1454
+        (context.response as any)._bodyInit) &&
```
pi0, line 199: We need to handle this line now since there is a second source:

https://github.com/joshmossas/ofetch/blob/e11a6a0ca98196c239bd3f65c92be847abdc6898/src/fetch.ts

joshmossas: @pi0 Okay I can make a quick patch

joshmossas: Although I should mention, that streams still don't work properly on react native so idk if this is actually necessary.

Their fetch implementation is kinda jank lol

joshmossas: Sure I can check real quick

pi0 (approved):

src/fetch.ts
pi0: Thanks! Merging in between feel free to amend method if had any new findings.

joshmossas:
@pi0 so I just checked with React Native running on my android phone

When `responseType` is set to `"stream"` the `_bodyInit` is an instance of `Blob`. However it doesn't implement any of the methods specified in the [Web standard](https://developer.mozilla.org/en-US/docs/Web/API/Blob) such as `_bodyInit.stream()` or `_bodyInit.text()`. It only implements the following interface:

```ts
interface Blob {
  readonly size: number;
  readonly type: string;
  slice(start?: number, end?: number, contentType?: string): Blob;
}
```

So none of the standard methods are present. As far as I can tell there's no way to get the actual data out of it.

joshmossas:
Additionally logging the _bodyInit to the console outputs the following:
```ts
{"_data": {"__collector": {}, "blobId": "244dfb46-0c73-45f8-99be-cfd8b2bb971f", "offset": 0, "size": 1256}}
```

When normally logging a Blob would look like this:
```ts
Blob { size: 11, type: '' }
```

pi0:
I see, basically seems stream mode is broken. 

Although now other types (text/json) work right?

joshmossas:
Yes I had no problem getting text and json to work

Issue #71, opened 2022-04-22
How to prevent custom headers transformed lowercase?


tsulatsitamim:
![image](https://user-images.githubusercontent.com/5946791/164600897-5538d640-54b4-47d3-80cd-876c34385841.png)
![2022-04-22_11-21](https://user-images.githubusercontent.com/5946791/164601373-69b4c5d8-c66c-4a74-954e-41080247fac0.png)

I try to send custom header, but it send in lowercase which my server doesn't accept it.

516310460:
nuxt3 issue: https://github.com/nuxt/framework/issues/6657

516310460:
The keys in the custom 'headers' will be wrapped by the' headers' object and automatically converted to lowercase

NozomuIkuta:
Closed according to https://github.com/nuxt/framework/issues/6657#issuecomment-1222238122

Please let us know if you find a provider that cares case-sensitivity.

PR #401, opened 2024-06-11
feat: allow custom global options for `$fetch.create`

CafuChino:
### 🔗 Linked issue

#71

### ❓ Type of change

### 📚 Description

This commit adds an optional parameter `globalOptionsOverwrite`  to  `$fetch.create`.  

However, for HTTP, headers are not case-sensitive. But for some reason, headers are case-sensitive in some old backend frameworks. This optional parameter overrides native `Headers` behaviour in such situations.

As only adding an optional parameter, it won`t break anything, maybe?

### 📝 Checklist

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -240,13 +240,13 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
 
   $fetch.native = (...args) => fetch(...args);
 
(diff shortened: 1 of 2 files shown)


pi0 (approved):
LGTM. Although modifying standard globals should be last resort, there are legitimate usecases and it is better than patching actual globals

PR #419, opened 2024-07-24
chore(types): provide types for payload methods

iiio2:
the types for payload methods are given. previously it was set of strings. now it is more specific. I think it might be good.


pi0:
thanks for the PR i really appreciate all your contributions.

i think i prefer to keep it as is because it is an internal utility and this change won't bring additional type safety but makes reading code a little bit harder.

iiio2:
ok @pi0, thank you.

PR #421, opened 2024-07-25
refactor: assign error response to variable

iiio2:
the PR is quite similar to this PR in Nitro (https://github.com/unjs/nitro/pull/2577) regarding `return await` in asynchronous function.


pi0:
Thanks for PR but it doesn't makes any different internally and while for documentation purposes we do this, making it internally only adds few more bytes.

iiio2:
Ok @pi0 . Thank you.

PR #423, opened 2024-07-27
chore: combine two conditions

iiio2:

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -8,7 +8,7 @@ export function isPayloadMethod(method = "GET") {
 }
 
 export function isJSONSerializable(value: any) {
-  if (value === undefined) {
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -8,7 +8,7 @@ export function isPayloadMethod(method = "GET") {
 }
 
 export function isJSONSerializable(value: any) {
-  if (value === undefined) {
+  if (value === undefined || value.buffer) {
```
pi0, line 11: This is unsafe before checking other falsy cases we do in L15 (false and null). If value is `null`, `value.buffer` makes an error

iiio2: I see. Thanks @pi0

PR #426, opened 2024-07-29
chore: add experimental for `Too Early` status

iiio2:
425 Too Early status is still experimental technology. I think may be adding experimental could be good for understanding.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Early-Data
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ await ofetch("/url", { ignoreResponseError: true });
 
 - `408` - Request Timeout
 - `409` - Conflict
-- `425` - Too Early (Experimental)
+- `425` - Too Early ([Experimental](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Early-Data))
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -121,7 +121,7 @@ await ofetch("/url", { ignoreResponseError: true });
 
 - `408` - Request Timeout
 - `409` - Conflict
-- `425` - Too Early
+- `425` - Too Early (Experimental)
```
pi0, line 124:
```suggestion
- `425` - Too Early ([Experimental](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Early-Data))
```

pi0 (approved):
thanks.

iiio2:
thank you @pi0 .

PR #428, opened 2024-08-11
perf: add typed arrays

iiio2:
Typed Arrays are added for `retryStatusCodes` and `nullBodyResponses`. I think may be typed arrays are considered more memory efficient.


pi0:
Thank you for the PR but i think in this case simplicity values more.

iiio2:
Ok @pi0 . Thank you.

PR #436, opened 2024-08-28
feat: always clone and normalize `options.headers` and `options.query`

### LINKED ISSUE #393: Inconsistent headers object returned after a retried request
### Environment

ofetch 1.3.4
node 20.12.2

### Reproduction

https://stackblitz.com/edit/vitejs-vite-53ww8b?file=main.js

Open the devtools console and click the POST REQUEST button to see the issue

You can see how the headers are being turned into a Headers object no matter the original input, making the headers not accesible anymore when the request loops back

### Describe the bug

coming from https://github.com/nuxt/nuxt/issues/27042

`$fetch` [is transforming the headers object](https://github.com/unjs/ofetch/blob/main/src/fetch.ts#L129) into a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers) class if there's a body and the method is other than GET; so if headers are being added in the `onRequest` interceptor using normal object property accessors instead of the Headers constructor (as shown in the [`useFetch` documentation](https://nuxt.com/docs/api/composables/use-fetch) at the moment I'm writing this issue), when the request is retried and loops back, these are being added as extra properties of the Headers object and ignored later

The expected behavior would be that the headers object was consistent across retried requests, and even across all interceptors

I think the headers object should either be normalized or at least be clear that the Headers constructor is required for modifying the request inside interceptors

### Additional context

I came across this issue when I tried to refresh a JWT using the `retry` option along with the `onResponseError` interceptor: https://github.com/nuxt/nuxt/discussions/26991

I also tried to fix it in this commit, normalizing the headers before the retry and after the onRequest: https://github.com/cmgdragon/ofetch/commit/8954d8c2f828004c6af703579eab422ce7404bdf, though this solution isn't taking into consideration any other interceptor but `onRequest` for modifying the headers

### Logs

_No response_

pi0:
resolves #393

This PR adds a change to always normalize `context.options.headers` into a `Headers` object and making sure it is also cloned. 

Additional fixes: 
- If first param to ofetch is a Request, and there is no headers option, it's headers will be used (same as fetch behavior)
- `query` and `params` are now consistent and normalized to the same object.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -98,6 +98,7 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
     const context: FetchContext = {
       request: _request,
       options: resolveFetchOptions<R, T>(
+        _request,
         _options,
         globalOptions.defaults as unknown as FetchOptions<R, T>,
         Headers
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -6,14 +6,17 @@ import {
   isPayloadMethod,
   isJSONSerializable,
   detectResponseType,
-  mergeFetchOptions,
+  resolveFetchOptions,
   callHooks,
 } from "./utils";
 import type {
   CreateFetchOptions,
   FetchResponse,
+  ResponseType,
   FetchContext,
   $Fetch,
+  FetchRequest,
+  FetchOptions,
 } from "./types";
 
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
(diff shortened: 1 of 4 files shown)


sondh0127:
this break the headers assign method. options.header.Authorization. please add breaking-change note

pi0:
It is not a breaking change since in < 1.4.0, the type of `context.options.headers` was `HeadersInit | Headers` (previous logic was supposed to handle all 3 possible values).

sondh0127:
> It is not a breaking change since in < 1.4.0, the type of `context.options.headers` was `HeadersInit | Headers` (previous logic in interceptors was supposed to handle all 3 possible values). (added note btw in [v1.4.0](https://github.com/unjs/ofetch/releases/tag/v1.4.0))

thanks for the update.

I didn't realize the header had three possible values

PR #440, opened 2024-09-05
refactor!: replace `ufo` with native `URL` utils

johannschopplich:
> [!NOTE]
> This is rather an idea and this PR might be better off over at [`ufo`](https://github.com/unjs/ufo). Feel free to close this PR if it should be handled otherwise.

Migrating `ufo` to use native `URL` [is already in discussion](https://github.com/unjs/ufo/issues/208), which would be beneficial to reduce the bundle size in the whole UnJS ecosystem, I suppose.

In @pi0 work on `h3` v2, `ufo` is no longer used and path utilities [have been internalized to the repo](https://github.com/unjs/h3/blob/main/src/utils/internal/path.ts). This might be also an approach for `ofetch`, since it only uses the following exports:
- `withBase`
- `withQuery`

Taking inspiration from the refactors in `h3` v2, both path utilities have been inlined to remove the dependency on `ofetch`. At the time of writing, one test still fails.

### DIFF
--- a/src/path.ts
+++ b/src/path.ts
@@ -1,0 +1,108 @@
+/* eslint-disable unicorn/prefer-at */
+export type QueryValue =
+  | string
+  | number
+  | boolean
+  | QueryValue[]
+  | Record<string, any>
+  | null
+  | undefined;
+export type QueryObject = Record<string, QueryValue | QueryValue[]>;
+
+/**
+ * Removes the leading slash from the given path if it has one.
+ */
+export function withoutLeadingSlash(path?: string): string {
+  if (!path || path === "/") {
+    return "/";
+  }
+
+  return path[0] === "/" ? path.slice(1) : path;
+}
+
+/**
+ * Removes the trailing slash from the given path if it has one.
+ */
+export function withoutTrailingSlash(path?: string): string {
+  if (!path || path === "/") {
+    return "/";
+  }
+
+  return path[path.length - 1] === "/" ? path.slice(0, -1) : path;
+}
+
+/**
+ * Joins the given base URL and path, ensuring that there is only one slash between them.
+ */
+export function joinURL(base?: string, path?: string): string {
+  if (!base || base === "/") {
+    return path || "/";
+  }
+
+  if (!path || path === "/") {
+    return base || "/";
+  }
+
+  const baseHasTrailing = base[base.length - 1] === "/";
+  const pathHasLeading = path[0] === "/";
+  if (baseHasTrailing && pathHasLeading) {
+    return base + path.slice(1);
+  }
+
+  if (!baseHasTrailing && !pathHasLeading) {
+    return `${base}/${path}`;
+  }
+
+  return base + path;
+}
+
+/**
+ * Adds the base path to the input path, if it is not already present.
+ */
+export function withBase(input = "", base = ""): string {
+  if (!base || base === "/") {
+    return input;
+  }
+
+  const _base = withoutTrailingSlash(base);
+  if (input.startsWith(_base)) {
+    return input;
+  }
+
+  return joinURL(_base, input);
+}
+
+/**
+ * Returns the URL with the given query parameters. If a query parameter is undefined, it is omitted.
+ */
+export function withQuery(input: string, query: QueryObject): string {
+  const url = new URL(input, "http://localhost");
+  const searchParams = new URLSearchParams(url.search);
+
+  for (const [key, value] of Object.entries(query)) {
+    if (value === undefined) {
+      searchParams.delete(key);
+    } else if (typeof value === "number" || typeof value === "boolean") {
+      searchParams.set(key, String(value));
+    } else if (!value) {
+      searchParams.set(key, "");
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for initiating pr dear @johannschopplich. Yes within unjs we are reducing dependency on ufo parser to adopt `URL`.

Considering wider ecosystem dependency and potentially hoisted deps, i do belive we should find a way to distribute double format (`ufo`/`ufo/legacy` with custom parser and `ufo/url` with native `URL` parser). (and i don't know how but this is the path i do think we should be taking.

src/path.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  return joinURL(_base, input);
+}
+
+/**
+ * Returns the URL with the given query parameters. If a query parameter is undefined, it is omitted.
+ */
+export function withQuery(input: string, query: QueryObject): string {
+  const url = new URL(input, "http://localhost");
```
pi0, line 79: Using URL parser can be expensive if input already has no params (which is case of most calls with bare URL + `query` object) we should avoid it and only parse the query section if query exists in input.

johannschopplich: Good catch! I've reduced `URL` usage in https://github.com/unjs/ofetch/pull/440/commits/997afbb0885542c7e4aec59882ac2df1fc34c0f2. This unintentionally also fixed the missing test. 😄

src/path.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Returns the URL with the given query parameters. If a query parameter is undefined, it is omitted.
+ */
+export function withQuery(input: string, query: QueryObject): string {
+  let url: URL | undefined;
+  let searchParams: URLSearchParams;
+
+  if (input.includes("?")) {
+    url = new URL(input, "http://localhost");
+    searchParams = new URLSearchParams(url.search);
```
pi0, line 84: We still don't need full `URL` to access search section we can use fast index check. ([see this](https://github.com/unjs/h3/blob/main/src/adapters/web/event.ts#L32-L79))

(includes above could be merged too with single string search)

src/path.ts
```diff
@@ -0,0 +1,119 @@
...
+  } else {
+    searchParams = new URLSearchParams();
+  }
+
+  for (const [key, value] of Object.entries(query)) {
```
pi0, line 89: We can also take a fast-path when there is no query in the original input. And simply use `URLSearchParams` as formatted

src/path.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+export function withQuery(input: string, query: QueryObject): string {
+  if (!query || Object.keys(query).length === 0) {
+    return input;
+  }
+
+  const searchIndex = input.indexOf("?");
+  const base = searchIndex === -1 ? input : input.slice(0, searchIndex);
+  const searchParams = new URLSearchParams(
+    searchIndex === -1 ? "" : input.slice(searchIndex + 1)
```
pi0, line 86: fast-path: when there is no searchIndex we can avoid creating new `URLSearchParams` (sorry for being strict this is kind of thing we forget later and it can be potentially bases to `ufo/lite` trying to make sure optimizing it as much as we can)

src/path.ts
johannschopplich: I think I get it. Is it better in the new code?

Another question in this regard: Would you prefer this one to omit the empty string in favour of `undefined`?

```diff
const searchParams = new URLSearchParams(
-  searchIndex === -1 ? "" : input.slice(searchIndex + 1)
+  searchIndex === -1 ? undefined : input.slice(searchIndex + 1)
);
```

By the way, I'm using the tests from `ufo` to check against this `withQuery` implementation.

src/path.ts
johannschopplich: Sorry, I don't get it completely: Even if no search is in the input path, we still need to construct the encoded query string for the new output path, am I right?
How should we build the search params if not using `URLSearchParams`? Port the `stringifyQuery` function from `ufo`? 

P.S.: Of course, no need to justify optimization and future-proofness. 🙋‍♂️

pi0: Sorry my wording was confusing. I meant that when there is no quesry in input string, we can have a fast path if block. So we don't need to iterate to for example remove `undefined` values, etc (we already know query state is empty) so it can be something like `return new URLSearchParams(query).toString` (I'm not sure if query object needs normalization but it could be simple map)

johannschopplich: Ahh, I get it now! Thanks for the clarification. Now:
1. If the query is empty, we return early.
2. If the input URL has no query, we only normalize the items of the input `query` object.

We have to normalize the values for `URLSearchParams`, otherwise it returns string like `key=undefined`.

I have again tested with the (slighty changed) `ufo` tests. I will create a separate PR there to track the implementation of these lighter functions.

TrySound:
What is left to complete this PR? Can I help?

pi0:
It is planned for next major version alongside other dep drops (like node-fetch-native and node < 20)

PR #447, opened 2024-09-29
refactor(fetch): Add method aliases for HTTP verbs

### LINKED ISSUE #282: Request method to function alias
### Describe the feature

Hi, is it possible to have the request method as a function alias on `ofetch` as it's on `axios` where we have something like
$fetch.get('url', {config})
$fetch.post('url', {config})

https://github.com/axios/axios#request-method-aliases

Kiansa:
This commit refactors the `fetch` function in the `src/fetch.ts` file to add method aliases for common HTTP verbs. The aliases include `get`, `post`, `put`, `delete`, `patch`, `head`, and `options`. These aliases simplify the usage of the `fetch` function by allowing developers to use more intuitive method names instead of manually specifying the HTTP method in the options object.

The changes also include updating the `types.ts` file to define the `FetchWithAliases` type, which extends the existing `$Fetch` type and adds the method aliases as properties.

This enhancement improves the readability and maintainability of the codebase by providing a more expressive and concise API for making HTTP requests. 

Closes #282

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -15,6 +15,7 @@ import type {
   ResponseType,
   FetchContext,
   $Fetch,
+  FetchWithAliases,
   FetchRequest,
   FetchOptions,
 } from "./types";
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for nice PR dear @Kiansa. 

I'm gonna some time to think about API and overhead but positive about DX benefits.

PR #462, opened 2024-11-22
feat: Add a isFetchError typeguard for FetchError

### LINKED ISSUE #460: Add a isFetchError typeguard for FetchError
### Describe the feature

Add a isFetchError typeguard for FetchError

### Additional information

Hugothms:
resolves #460


pi0:
I appreciate the PR. Depending on `instanceof` can be unstable assumption specially if there might be multiple versions of ofetch (bundled, version, etc) it gives wrong result.

We can do a Symbol or prop check but I alternatively suggest that we assign a specific error `.code` such as `OFETCH_ERROR` so can be checked externally. PR welcome for this 👍🏼

PR #482, opened 2025-03-28
feat: serialize with `URLSearchParams` for `application/x-www-form-urlencoded` content type header

### LINKED ISSUE #425: Check if the body should be parsed to handle x-www-form-urlencoded payloads
### Describe the feature

`application/x-www-form-urlencoded` is also very common. If $fetch can resolve it automatically, it will be functional.

### Additional information

kricsleo:
…orm-urlencoded`

resolves #425


pi0 (approved):
Thanks ❤️

PR #483, opened 2025-03-29
fix: mark `FormData` & `URLSearchParams` as non-serializable for bun compatibility

### LINKED ISSUE #439: Bun compatibility
### Describe the feature

I tried `ofetch` with Bun. It seems to work, but I ran into a small issue when POSTing `multipart/form-data`.  It seems the Bun `FormData` class has a `toJSON` method, while Node and Deno do not. It causes the `isJSONSerializable()` method to return `true` (exact line below), and therefore sends the serialized payload as `application/json` instead of `multipart/form-data`.

Would a `value instanceof FormData` check make sense here?

https://github.com/unjs/ofetch/blob/7726adbf10a5a233ae98cb8adb619864b012db65/src/utils.ts#L36

### Additional information

- [ ] Would you be willing to help implement this feature?

### LINKED ISSUE #466: POST request with x-www-form-urlencoded is failing as Bad Request
### Environment

bun: 1.1.38
node: 18.17.0
ofetch: 1.4.1

### Reproduction

I have a simple api that accepts a name and return a json response.
Steps:
1.  GO api to handle the request
```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
	"net/http"
)

func main() {
	http.HandleFunc("/hello", helloHandler)
	fmt.Println("Server is running on port 8080...")
	log.Fatal(http.ListenAndServe(":8080", nil))
}

func helloHandler(w http.ResponseWriter, r *http.Request) {
	if r.Method != http.MethodPost {
		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
		return
	}

	err := r.ParseForm()
	if err != nil {
		http.Error(w, "Unable to parse form data", http.StatusBadRequest)
		return
	}

	name := r.FormValue("name")
	if name == "" {
		http.Error(w, "Name field is required", http.StatusBadRequest)
		return
	}

	response := map[string]string{"hello": name}

	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)

	json.NewEncoder(w).Encode(response)
}
```
2. Run via `go run app.go`
3. Try using the endpoint via
```ts
import { ofetch } from "ofetch";

const options = {
    method: 'POST',
    headers: {'Content-Type': 'application/x-www-form-urlencoded'},
    body: new URLSearchParams({
      name: 'abc'
    })
  };
  
  ofetch('http://localhost:8080/hello', options)
    .then(response => console.log(response))
    .catch(err => console.error(err));
```

It should error out.

### Describe the bug

Form encoded post requests are not working as expected and erroring out(400).

### Additional context

I tried using simple fetch and things were working as expected.
```js
const options = {
    method: 'POST',
    headers: {'Content-Type': 'application/x-www-form-urlencoded'},
    body: new URLSearchParams({
      name: 'abc'
    })
  };
  
  fetch('http://localhost:8080/hello', options)
    .then(response => response.json())
    .then(response => console.log(response))
    .catch(err => console.error(err));
```

Even curl was working:
```bash
curl --request POST http://localhost:8080/hello --header 'Content-Type: application/x-www-form-urlencoded' --data name=abc
```

### Logs

```sh
FetchError: [POST] "http://localhost:8080/hello": 400 Bad Request
```

kricsleo:
resolves #439
resolves #466

Bun added `toJSON` method to `URLSearchParams` and `FormData`, this confused the `isJSONSerializable`.

It’s neither Node.js behavior (Bun aims for 100% compatibility) nor standard behavior (according to MDN)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,6 +31,12 @@ export function isJSONSerializable(value: any) {
   if (value.buffer) {
     return false;
   }
+  // FormData should't be JSON serializable,
+  // but `Bun` adds a `toJSON` method to it, which is not standard.
+  // issue: https://github.com/unjs/ofetch/issues/439
+  if(value instanceof FormData) {
+    return false;
+  }
   return (
     (value.constructor && value.constructor.name === "Object") ||
     typeof value.toJSON === "function"
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -34,7 +34,7 @@ export function isJSONSerializable(value: any) {
   // FormData should't be JSON serializable,
   // but `Bun` adds a `toJSON` method to it, which is not standard.
   // issue: https://github.com/unjs/ofetch/issues/439
-  if(value instanceof FormData) {
+  if (value instanceof FormData) {
     return false;
   }
   return (
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,10 +31,9 @@ export function isJSONSerializable(value: any) {
   if (value.buffer) {
     return false;
   }
-  // FormData should't be JSON serializable,
-  // but `Bun` adds a `toJSON` method to it, which is not standard.
-  // issue: https://github.com/unjs/ofetch/issues/439
-  if (value instanceof FormData) {
+  // `FormData` and `URLSearchParams` should't have a `toJSON` method,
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks ❤️

PR #486, opened 2025-04-02
feat: auto detect `text/event-stream` as `stream` response type

### LINKED ISSUE #484: unable to handle SSE(server-sent events)
### Environment

ofetch: 1.4.1
Node.js: N/A

### Reproduction

```ts
import { ofetch } from 'ofetch'
const resp = await ofetch("https://echo.epb.wiki/_sse", {
	query: { timeout: 3 },
});
```

### Describe the bug

The above code stuck for 3 seconds, but I expect it to return immediately, 
and allowing me to process every individual events.

May be adding a hook between send the request and parsing body,
or have some ways to customize body parsing.

### Additional context

_No response_

### Logs

```sh

```

kricsleo:
resolves #484

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ const { users } = await ofetch("/api/users");
 
 For binary content types, `ofetch` will instead return a `Blob` object.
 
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

PR #487, opened 2025-04-02
docs: guide on augmenting `FetchOptions`

### LINKED ISSUE #103: Extend options to add additional metadata?
Sometimes I would like to handle certain responses for requests in a certain way, such as a 422 returning pop up notifications with the errors. But I don't want to do this for ALL requests, so was wondering if there was  a way to pass additional options such as "metadata" which can then be checked.

```js
const useAPIFetch = $fetch.create({
  async onResponseError({ request, response, options }) {
    if(options.metadata.showPopUps) {
      // show popup
    }
  },
})
```

kricsleo:
resolve #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,36 @@ const apiFetch = ofetch.create({ baseURL: "/api" });
 apiFetch("/test"); // Same as ofetch('/test', { baseURL: '/api' })
 ```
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks ❤️

PR #490, opened 2025-04-03
feat: allow passing `URLSearchParams` to `query` config

### LINKED ISSUE #322: Support usage of URLSearchParams as a value for query / params
### Describe the feature

I naively used `URLSearchParams` in order to prepare the parameters I want to send along the request, however those are not read if I use an interceptor later on.

```ts
const $fetch = ofetch.create({
  baseURL: this.baseUrl,
  onRequest: ({ options }) => {
    options.query ||= new URLSearchParams();
    options.query.set('token', this.token);
  }
});

const query = new URLSearchParams({
  title,
});

$fetch(url, { query })
```

Is it something unwanted, or is a contribution welcome for this one ?

### Additional information

ysknsid25:
resolves https://github.com/unjs/ofetch/issues/322


src/types.ts
```diff
@@ -32,6 +32,8 @@ export interface FetchOptions<R extends ResponseType = ResponseType, T = any>
 
   query?: Record<string, any>;
```
pi0, line 33: I think we can add this feature to same `query` option to support both

ysknsid25: Thank you for pointing out ! I reflect your feedback.

PR #498, opened 2025-05-10
ci: install corepack globally

iiio2:


pi0:
CI seems to be ok.

iiio2:
Ok @pi0 . Thanks.

PR #499, opened 2025-05-10
chore: provide `https` instead of `http`

iiio2:
Here I think we can provide `https`. It works with `http`, but in the google chrome browser it shows not secure. That's why I think may be we can give `https` in those links.

### DIFF
--- a/playground/index.mjs
+++ b/playground/index.mjs
@@ -1,7 +1,7 @@
 import { $fetch } from "..";
 
 async function main() {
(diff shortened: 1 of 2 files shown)


pi0:
thanks for PR but testing http=>https redirect was goal of this use! i might randomly change them later but it is not needed!

iiio2:
I see. Ok, @pi0 . Thanks.

Issue #326, opened 2023-11-19
Support custom Abortcontroller with timeout


pi0:
When `timeout` option is provded, we need to set a custom abort controller. This means using `timeout` and `controller` are mutually exclusive. ([src](https://github.com/unjs/ofetch/blob/bb2d72baa5d3f332a2185c20fc04e35d2c3e258d/src/fetch.ts#L152)).

I am thinking if there is (low cost) solution to merge signals

PR #508, opened 2025-08-04
feat: support custom `AbortSignal` with `timeout`

OrbisK:
This PR uses a `AbortSignal.timeout()` merged with an optional `options.signal` with `AbortSignal.any()`

resolves https://github.com/unjs/ofetch/issues/326

This should make the following obsolete: https://github.com/unjs/ofetch/pull/481.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -35,11 +35,8 @@ const retryStatusCodes = new Set([
 const nullBodyResponses = new Set([101, 204, 205, 304]);
 
 export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
-  const {
-    fetch = globalThis.fetch,
-    Headers = globalThis.Headers,
-    AbortController = globalThis.AbortController,
-  } = globalOptions;
+  const { fetch = globalThis.fetch, Headers = globalThis.Headers } =
+    globalOptions;
 
   async function onError(context: FetchContext): Promise<FetchResponse<any>> {
     // Is Abort
@@ -166,12 +163,17 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
       }
     }
 
-    const mergedSignal = AbortSignal.any([
-      typeof context.options.timeout === "number" ? AbortSignal.timeout(context.options.timeout) : undefined,
-      context.options.signal,
-    ].filter((s): s is NonNullable<typeof s> => Boolean(s)))
+    const mergedSignal = AbortSignal.any(
+      [
+        typeof context.options.timeout === "number"
+          ? AbortSignal.timeout(context.options.timeout)
+          : undefined,
+        context.options.signal,
+        // eslint-disable-next-line unicorn/prefer-native-coercion-functions
+      ].filter((s): s is NonNullable<typeof s> => Boolean(s))
+    );
 
-    context.options.signal = mergedSignal
+    context.options.signal = mergedSignal;
 
     try {
       context.response = await fetch(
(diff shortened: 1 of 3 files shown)


kricsleo:
I'm not using these APIs just for compatibility reasons; `AbortSignal.timeout()` is available in [Chrome 124+](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static#browser_compatibility), and `AbortSignal.any` is available in [Node.js 20.3+](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static#browser_compatibility). This seems too strict.

OrbisK:
like this?

src/fetch.ts
```diff
@@ -165,21 +162,14 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
...
-        controller.abort(error);
-      }, context.options.timeout);
-      context.options.signal = controller.signal;
+    if (typeof context.options.timeout === "number") {
+      context.options.signal = AbortSignal.any(
+        [
+          AbortSignal.timeout(context.options.timeout),
+          context.options.signal,
+          // eslint-disable-next-line unicorn/prefer-native-coercion-functions
+        ].filter((s): s is NonNullable<typeof s> => Boolean(s))
```
pi0, line 171: We need `AbortSignal.any` wrapper if the array length is more than 1, right?

src/fetch.ts
```diff
@@ -165,21 +162,14 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
...
-        );
-        error.name = "TimeoutError";
-        (error as any).code = 23; // DOMException.TIMEOUT_ERR
-        controller.abort(error);
-      }, context.options.timeout);
-      context.options.signal = controller.signal;
+    if (typeof context.options.timeout === "number") {
+      context.options.signal = AbortSignal.any(
+        [
+          AbortSignal.timeout(context.options.timeout),
```
pi0, line 168: Moving from `setTimeout` to `AbortSignal.timeout` is a new requirement (**even** if it is supported in baseline browser versions) we need to use `setTimeout` when there is no custom `context.options.signal`

OrbisK: Ohh. So if someone passes a signal, we know signals are supported? I can adjust this :)

src/fetch.ts
```diff
@@ -168,18 +168,27 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
...
-        const error = new Error(
-          "[TimeoutError]: The operation was aborted due to timeout"
+    if (context.options.timeout) {
+      if (context.options.signal) {
+        context.options.signal = AbortSignal.any(
+          [
+            AbortSignal.timeout(context.options.timeout),
+            context.options.signal,
+            // eslint-disable-next-line unicorn/prefer-native-coercion-functions
+          ].filter((s): s is NonNullable<typeof s> => Boolean(s))
```
pi0, line 178: We already checked in L171 and L172 that both are non nullable i think filter is extra here.

src/fetch.ts
```diff
@@ -168,18 +168,27 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
...
         );
-        error.name = "TimeoutError";
-        (error as any).code = 23; // DOMException.TIMEOUT_ERR
-        controller.abort(error);
-      }, context.options.timeout);
-      context.options.signal = controller.signal;
+      } else {
+        const controller = new AbortController();
```
pi0, line 181:
```suggestion
        // TODO: Use AbortSignal.timeout in next major
        const controller = new AbortController();
```

Issue #522, opened 2025-10-31
context.options.headers.get is not a function in `Nuxt` 4.2.0 with `ofetch` 1.5.0


Flo0806:
### Environment

Since the latest version `1.5.0` `ofetch` throws an error if `$fetch` it used in `Nuxt 4.2.0`.

Linked issue: https://github.com/nuxt/nuxt/issues/33604#issuecomment-3471152141

Reproduction: 
> I've made small reproduction of this issue
https://stackblitz.com/edit/github-kbqwxvpo?file=app%2Fapp.vue
In App.vue file, I create new $fetch client and tries to send POST request with body and in console, you will see the error

### Reproduction

```Vue
<script setup lang="ts">
const httpOptions = {
  baseURL: '/api',
  credentials: 'include',
  headers: {
    Accept: 'application/json',
  },
  retry: false,

  async onRequest({ options, request }) {
    options.headers = {
      Accept: 'application/json',
    };
    console.log(options.headers);
  },
};

const client = $fetch.create(httpOptions);

const test = async () => {
  console.log('SEND POST REQUEST');
  client('/login', {
    method: 'post',
    body: {
      email: 'test@mail.com',
      password: '12312313',
    },
  });
};
</script>

<template>
  <div>
    <button @click="test">click</button>
  </div>
</template>

<style scoped></style>
```

This code will log a browser error: `context.options.headers.get is not a function`

### Describe the bug

It throws the error `context.options.headers.get is not a function` if custom headers are set when using `$fetch` in a `Nuxt 4.2.0` project.

### Logs

```sh

```

Issue #521, opened 2025-10-29
context.options.headers.get is not a function


dongnaebi:
### Environment

all environment

### Reproduction

```ts
export const $ofetch = ofetch.create({
  timeout: 15000,
  onRequest({ options }) {
    options.headers = {
      ...options.headers,
      Authorization: token.value,
      UUID: uuid.value
    } as any
  },
})
```

### Describe the bug

```
context.options.headers.get is not a function
```

I started getting this error after upgrading to version 1.5.0 — everything worked fine in 1.4.1 and earlier versions.

### Logs

```sh

```

kricsleo:
Can't reproduce. Could you provide a minimal reproduction using https://stackblitz.com/?

Umid-Berdiev:
got TypeError: u.options.headers.get is not a function in NuxtJS app

kricsleo:
> Can't reproduce in `v1.5.0`. Could you provide a minimal reproduction with [this example](https://stackblitz.com/~/github.com/kricsleo/starter-ts-minimal?file=index.mjs&startScript=dev)? 🙏

Thanks for all the reports, but could anyone provide a reproduction? 🙏

yaowangmx:
> nuxt@3.18.1 too

Solution: Use `new Headers` to set the request headers in `onRequest`.
For example:
```
onRequest({ options }) {
  if (process.env.NODE_ENV === "production") {
    if (process.client) {
      options.baseURL = config.public.baseURL
    } else {
      options.baseURL = config.public.baseIntrantURL
    }
  }
  const hdrs = new Headers(options.headers || {})
  
  hdrs.set("X-Platform", "web")

  if (!process.client) {
    const reqHeaders = useRequestHeaders(["cookie"]) || {}
    if (reqHeaders.cookie) {
      hdrs.set("cookie", reqHeaders.cookie)
    }
    const { locale } = useI18n()
    if (locale?.value) {
      hdrs.set("locale", locale.value)
    }
  }

  options.headers = hdrs
}
```

PR #524, opened 2025-10-31
fix: normalize `options.headers` (again) after `onRequest` hook

kricsleo:
resolves https://github.com/unjs/ofetch/issues/522
(resolves https://github.com/unjs/ofetch/issues/521)

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -114,18 +114,10 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
 
     if (context.options.onRequest) {
       await callHooks(context, context.options.onRequest);
-
       if (!(context.options.headers instanceof Headers)) {
-        const receivedType =
-          (context.options.headers as unknown)?.constructor?.name ||
-          typeof context.options.headers;
-        console.warn(
-          "[ofetch]: Invalid type for `options.headers`. " +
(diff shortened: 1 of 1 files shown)


test/index.test.ts
```diff
@@ -468,6 +468,23 @@ describe("ofetch", () => {
...
 
+  it("hook modifications", async () => {
+    // onRequest
+    expect(
+      await $fetch(getURL("/echo"), {
+        method: "POST",
+        body: { num: 42 },
+        onRequest(ctx) {
+          ctx.options.headers = {
+            ...ctx.options.headers,
```
pi0: This usage is totally wrong. `ctx.options.headers` is typed explicitly as `Headers`

src/fetch.ts
pi0: We could alternatively move normalization to here + warn that hooks should not assign an object to `options.headers` (something real is broken probably as they omit all headers)

test/index.test.ts
kricsleo: Yeah, it should be. 
Previously, it was implicitly allowed because the internals were normalized as `Headers` regarding the type. A small change in `v1.5.0` breaks this implicit behavior. Although we made no promise regarding this, I think it would be better if we could keep that without extra burden (or bumping major versions).

src/fetch.ts
kricsleo: That's a nice idea. 👍
Updated.

pi0 (approved):

PR #545, opened 2026-03-17
fix: trim trailing whitespace from baseURL in withBase

### LINKED ISSUE #530: Merging URL Components
### Environment

Runtime: Browser
Browser Version: Google Chrome: Version 142.0.7444.175 (Official Build) (64-bit)

### Reproduction

```typescript
import { ofetch } from "ofetch";

ofetch('/path', {
  baseURL: 'https://example.org/',
  query: {
    q: 'y',
  },
  onRequest: [({ request }) => {
    console.log(request.toString()); // https://example.org/path?q=y
  }]
});

ofetch('/path', {
  baseURL: 'https://example.org/\r',
  query: {
    q: 'y',
  },
  onRequest: [({ request }) => {
    console.log(request.toString()); // https://example.org/?q=y
  }]
});

```

### Describe the bug

Merging URL components doesn’t work correctly in some scenarios. For example, when the baseURL ends with `\r` (e.g. `'https://example.org/\r'`), ofetch completely skips the path part and only appends the query parameters to the base URL. However, if the request has no query parameters, it still works, even with the `\r` at the end.

### Additional context

The `\r` ending character was a .env parsing mistake.

### Logs

```sh

```

guoyangzhen:
When `baseURL` ends with whitespace or control characters (e.g. `\r`, `\n`, `\t`) — which can happen from `.env` parsing mistakes — the URL merging breaks because `withoutTrailingSlash` only checks for `/` at the end.

## Root Cause

The `withBase()` function passes `base` to `withoutTrailingSlash()`, which only strips trailing `/`. When the base URL has trailing control characters like `\r`:

```
baseURL: "https://example.org/\r"
↓ withoutTrailingSlash → "https://example.org/\r" (unchanged, no / at end)
↓ joinURL("https://example.org/\r", "/path")
↓ Result: "https://example.org/\r/path" (invalid URL - path gets lost)
```

The browser/URL parser interprets `\r` as a line break, causing the path component to be stripped.

## Fix

Trim trailing whitespace from the base URL before processing in `withBase()`.

## Testing

Added tests for `\r\n`, `\t`, and trailing spaces in `baseURL`.

Fixes #530

## Summary by CodeRabbit

* **Chores**
  * Removed URL construction utilities module and related type definitions.

* **Tests**
  * Added validation for baseURL whitespace trimming across various whitespace characters.


src/utils.url.ts
Yizack, line 1: Btw did you accidentally delete the entire utils file?

PR #550, opened 2026-03-30
feat: merge hooks from defaults and per-request options

### LINKED ISSUE #319: Allow having interceptos from both defaults and per request

productdevbook:
## Summary

- Hooks from `$fetch.create()` defaults and per-request options are now **concatenated** instead of overwritten
- Default hooks run first, followed by per-request hooks
- Applies to all hook types: `onRequest`, `onRequestError`, `onResponse`, `onResponseError`

## Problem

Currently, per-request hooks completely replace default hooks set via `$fetch.create()`:

```ts
const api = $fetch.create({
  onRequest() { /* logging - this gets lost! */ }
})

api('/endpoint', {
  onRequest() { /* transform */ }
})
```

## Solution

Hooks are now merged using array concatenation (same pattern as `got.extend()`):

```ts
const api = $fetch.create({
  onRequest() { console.log('log') }  // runs first
})

api('/endpoint', {
  onRequest() { console.log('transform') }  // runs second
})
// Output: "log", "transform"
```

Array hooks also merge correctly:

```ts
const api = $fetch.create({
  onRequest: [hookA, hookB]
})

api('/endpoint', {
  onRequest: [hookC, hookD]
})
// Execution order: hookA → hookB → hookC → hookD
```

## Test plan

- [x] Default + per-request single hooks merge in correct order
- [x] Default + per-request array hooks merge in correct order
- [x] Works when only defaults have hooks
- [x] Works when only per-request has hooks
- [x] All hook types merge (`onResponse`, `onResponseError`, etc.)
- [x] All existing tests pass
- [x] Lint, typecheck, build pass

Resolves #319

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Request lifecycle hooks now merge when configured in both default and per-request settings. Hook handlers execute sequentially—defaults run first, followed by per-request hooks—rather than per-request hooks replacing defaults entirely. This enhancement applies to all lifecycle hooks: onRequest, onResponse, onRequestError, and onResponseError.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,7 @@
 import type {
   FetchContext,
   FetchHook,
+  FetchHooks,
   FetchOptions,
   FetchRequest,
   ResolvedFetchOptions,
@@ -109,13 +110,49 @@ export function resolveFetchOptions<
     };
   }
 
+  // Merge hooks: concatenate defaults and input hooks (defaults run first)
+  const mergedHooks = mergeHooks(
+    defaults as FetchOptions | undefined,
+    input as FetchOptions | undefined
+  );
+
   return {
     ...defaults,
     ...input,
+    ...mergedHooks,
     query,
     params: query,
     headers,
   };
+}
+
+const hookNames: Array<keyof FetchHooks> = [
+  "onRequest",
+  "onRequestError",
+  "onResponse",
+  "onResponseError",
+];
+
+function toArray<T>(value: T | T[] | undefined): T[] {
+  if (!value) {
+    return [];
+  }
+  return Array.isArray(value) ? value : [value];
+}
+
+function mergeHooks(
+  defaults: FetchOptions | undefined,
+  input: FetchOptions | undefined
+): Partial<FetchHooks> {
+  const merged: Partial<FetchHooks> = {};
+  for (const name of hookNames) {
+    const defaultHooks = toArray((defaults as any)?.[name]);
+    const inputHooks = toArray((input as any)?.[name]);
+    if (defaultHooks.length > 0 || inputHooks.length > 0) {
+      (merged as any)[name] = [...defaultHooks, ...inputHooks];
+    }
+  }
+  return merged;
 }
 
 function mergeHeaders(
--- a/test/index.test.ts
+++ b/test/index.test.ts
@@ -505,6 +505,89 @@ describe("ofetch", () => {
     expect(onResponseError).toHaveBeenCalledTimes(2);
   });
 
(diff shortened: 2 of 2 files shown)


maximepvrt:
@productdevbook thank you so much for working on this. This is going to be a major game changer. On the other hand, the approach proposed here seems interesting: https://github.com/unjs/ofetch/issues/319#issuecomment-2283394589

It allows for blocking the execution of previous ones if necessary and helps in organizing the order of the hooks.

PR #563, opened 2026-04-04
fix(timeout): add AbortSignal.timeout fallback for older Node.js versions (#532)

### LINKED ISSUE #532: In server side, the timeout setting is invalid, and the request cannot be aborted after the timeout.
### Environment

node v22.14.0
Nuxt 3.16.0
ofetch 1.4.1

### Reproduction

none

### Describe the bug

In server side, the timeout setting is invalid, and the request cannot be aborted after the timeout.
In client side,  it works well.

### Additional context

_No response_

### Logs

```sh

```

gkoos:
Resolves #532

## Problem

Timeout setting was invalid on server-side. In Node.js versions without `AbortSignal.timeout()` support (pre-17.3.0) or when unreliable, requests would not abort after timeout. The code declared an unused `abortTimeout` variable but never set it.

## Solution

Implement a fallback mechanism:

- Modern Node.js 17.3.0+: use native `AbortSignal.timeout()`
- Older Node.js / server environments: fall back to manual `setTimeout()`
- Cleanup: always clear timeout in finally block

## Changes
- `fetch.ts`: Add `AbortSignal.timeout` feature detection with fallback implementation
- `index.test.ts`: Add three timeout tests (abort, normal requests, cleanup)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling to be more compatible across different runtimes, ensuring timeouts are reliably enforced and cleaned up.

* **Tests**
  * Added tests for timeout rejection, fallback timeout behavior, successful responses within timeout windows, and concurrent request handling to prevent timer leakage.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -182,12 +182,35 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
         // Fallback for older Node.js / server environments
         abortController = new AbortController();
         abortTimeout = setTimeout(() => {
-          abortController?.abort();
+          abortController?.abort(
+            new DOMException("signal timed out", "TimeoutError")
+          );
         }, context.options.timeout);
 
-        context.options.signal = context.options.signal
-          ? AbortSignal.any([abortController.signal, context.options.signal])
-          : abortController.signal;
+        if (context.options.signal) {
+          if (typeof AbortSignal.any === "function") {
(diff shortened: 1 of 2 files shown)


pi0:
linked PR is using older version of ofetch you are making against main. Also minimum supported node version is 20+

PR #513, opened 2025-09-19
Add retryCondition Option

faizanu94:
Resolves #503

## Summary

This PR adds support for `retryCondition` option for `$fetch`.

### What’s new
- Introduces `retryCondition(context)` callback for custom retry logic.
- Keeps `retryStatusCodes` and `retryCondition` independent — retries if **either** is satisfied.
- If `retryCondition` throws, the error is propagated to the consumer.
- Functionality is verified via unit tests

## Example

```ts
await $fetch('/api/data', {
  retry: 3,
  retryCondition: (ctx) => {
    // Retry until response body contains expected field
    return ctx.response?._data?.status !== 'ready';
  }
});

PR #618, opened 2026-07-29
feat!: add support for manual retries from hooks

### LINKED ISSUE #355: Ability to retry a request after an error

### LINKED ISSUE #358: feat: add `retryCb` option

### LINKED ISSUE #503: Custom retry logic
### Describe the feature

Having a retry on specific status codes is great but could we get a more customizable way of getting a retry to happen? I think this is a common use case to retry a request if the response doest match criteria. 

Now I need to do custom wrapper logic to get such functionality.

```ts
//custom wrapper
export async function retryCall<T>({
  callback,
  maxRetries = 5,
  isResponseMatch,
  delayMs = 1000,
}: {
  maxRetries?: number;
  isResponseMatch: (data: T) => boolean;
  callback: () => Promise<T>;
  delayMs?: number;
}): Promise<T> {
  for (let attempt = 1; attempt < maxRetries; attempt++) {
    try {
      const response = await callback();

      if (isResponseMatch(response)) {
        return response;
      }
    } catch (error) {
      throw error;
    }

    await new Promise((resolve) => setTimeout(resolve, delayMs));
  }

  return await callback();
}
```

```ts
await retryCall({
      callback: () => ofetch(...),
      isResponseMatch: (data) => data === 'success',
    })
```

### Additional information

- [ ] Would you be willing to help implement this feature?

### LINKED ISSUE #536: Add `retryAttempt` to `FetchContext`
### Describe the feature

PR #372 added the ability to pass a callback to `retryDelay`, which should enable exponential backoff, but currently it's tricky to implement. I couldn't find an option to get the **current attempt** in this callback. The mentioned PR calls for `context.options.retryAttempt`, which doesn't exist.

I am currently using a workaround:
```ts
function exponentialBackoff(baseRetries: number, baseDelay = 1_000) {
    return (context: FetchContext<any, 'json'>) => {
      const retryLeft = context.options.retry || 0
      const currentRetry = baseRetries - retryLeft
      return baseDelay * 2 ** currentRetry
    }
  }
```

### Additional information

- [x] Would you be willing to help implement this feature?

### LINKED ISSUE #581: Exponential backoff with jitter for retries
# Summary

Add a new `retryBackoff` option that lets users configure exponential backoff with jitter for automatic retries, following the strategies described in the AWS Architecture Blog post ["Exponential Backoff And Jitter"](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) by Marc Brooker.

# Motivation

The current retry mechanism (`src/fetch.ts`) only supports a fixed `retryDelay` (a number or a callback returning a number). This is fine for a single client, but it scales poorly when many clients fail simultaneously:

1. With no delay (`retryDelay: 0`, the default), every client retries immediately and stampedes the server while it is still recovering.
2. With a fixed delay (e.g. `500ms`), all failed clients still retry in lockstep. the herd is preserved, just shifted in time. The server sees the same pulsing load pattern.

This is the classic "Thundering Herd" problem. The AWS post shows via simulation that adding exponential backoff alone is not enough; the key insight is that **randomizing each client's retry schedule (jitter)** is what actually decorrelates the herd. With jitter, both the peak load on the server and the overall completion time across all clients improve significantly.

Today, users who want this behavior have to implement it themselves inside a `retryDelay` callback, which is error-prone (need to track attempt count, pick the right formula, manage `prev_sleep` for decorrelated jitter, etc.). It would be much nicer if `ofetch` provided this out of the box.

# Current behavior

```ts
// src/types.ts
retry?: number | false;
retryDelay?: number | ((context: FetchContext<T, R>) => number);
retryStatusCodes?: number[];
```

```ts
// src/fetch.ts (onError)
const retryDelay =
  typeof context.options.retryDelay === "function"
    ? context.options.retryDelay(context)
    : context.options.retryDelay || 0;
```

The `FetchContext` does not expose the current attempt number, so even a user-defined callback cannot easily implement backoff without external state.

# Proposal

Introduce a new opt-in option `retryBackoff` that selects a backoff strategy and supplies `base` / `cap` delays in milliseconds:

```ts
await ofetch("/api", {
  retry: 5,
  retryBackoff: {
    strategy: "full-jitter", // "full-jitter" | "equal-jitter" | "decorrelated-jitter"
    base: 100,  // minimum delay (ms)
    cap: 3000,  // maximum delay (ms)
  },
});
```

## Strategies

All three strategies from the AWS post would be supported so users can pick the one that matches their constraints:

- **`full-jitter`** (recommended in the AWS post):
  `sleep = random(0, min(cap, base * 2 ** attempt))`
  Best overall — lowest server load and lowest completion time in the simulation. No state required.

- **`equal-jitter`**:
  `temp = min(cap, base * 2 ** attempt); sleep = temp / 2 + random(0, temp / 2)`
  Guarantees a minimum wait time (`temp / 2`), useful when "retry immediately" is undesirable.

- **`decorrelated-jitter`**:
  `sleep = min(cap, random(base, prev_sleep * 3))` (with `prev_sleep = base` on the first retry)
  Comparable to full-jitter in performance and matches the AWS SDK's default. Requires tracking the previous sleep value.

## Expose attempt count on the context

To keep callback-based use cases working and to let users observe retries, add a public `retryAttempt?: number` field to `FetchContext`:

- `undefined` on the first attempt
- `1`, `2`, ... on subsequent retries

## Coexistence with `retryDelay`

When both `retryBackoff` and `retryDelay` are set, `retryBackoff` takes precedence and `retryDelay` is ignored. Documented explicitly to avoid silent mixing of the two semantics.

## Backward compatibility

Fully opt-in. If `retryBackoff` is not specified, the existing `retryDelay` behavior is preserved exactly. No default delay is applied, no public type changes are breaking.

# References

- [AWS Architecture Blog: Exponential Backoff And Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)  Marc Brooker
- Existing retry code: `src/fetch.ts` (onError), `src/types.ts` (retry options), `README.md` ("Auto Retry" section)

### Additional information

- [x] Would you be willing to help implement this feature?

cernymatej:
resolves #581
resolves #503 (PR https://github.com/unjs/ofetch/pull/513)
resolves #358 (PR https://github.com/unjs/ofetch/pull/359)
resolves #355
resolves #536 (PR https://github.com/unjs/ofetch/pull/537)

> [!IMPORTANT]
> This PR is meant for the v2

This PR aims to resolve the above-mentioned issues by introducing a way to retry the current request based on custom logic in the interceptors. The PRs that are open are IMO not flexible enough to solve all the issues.

Interceptors can now start a retry by returning `ctx.retry()`:

```ts
const api = $fetch.create({
  async onResponseError(ctx) {
    if (ctx.response.status === 401 && ctx.retries.count("auth") === 0) {
      await refreshSession()
      return ctx.retry({
        cause: "auth",
        options: { headers: { /* ... */ } },
      })
    }
  },
})
```

The replayed request goes through the whole pipeline again, so all interceptors re-run.

Every retry is recorded in `ctx.retries`:

```ts
ctx.retries.count("auth") // retries with a given cause
ctx.retries.count() // all retries
ctx.retries.last // { cause, trigger, status? }
ctx.retries.history // readonly RetryEntry[]
```

### Merging / Cancelling

```ts
onResponseError: [
  // one layer starts the retry
  (ctx) => {
    if (isColdStart(ctx.response) && ctx.retries.count("warmup") < 2) {
      return ctx.retry({ cause: "warmup", delay: 1000 })
    }
  },
  // another one decorates any pending retry
  (ctx) => {
    if (ctx.pendingRetry) {
      return ctx.retry({
        delay: 3000, // longest delay wins
        options: { headers: { "x-retried-for": ctx.pendingRetry.cause! } },
      })
    }
  },
  // and a last one can still call it off
  (ctx) => {
    if (ctx.retries.count() >= 5) {
      ctx.cancelRetry()  // doesn't have to be returned
    }
  },
]
```

- `cause` is immutable once set
- headers merge per key
- `delay` coerces to the longest
- a retry targeting a different `request` than an already `ctx.pendingRetry` is ignored with an error log
- `ctx.retry()` throws for stream bodies

### Automatic retries

The declarative options (`retry`, `retryDelay`, `retryStatusCodes`) are unchanged, but the "budget" (`retry`, `retryDelay`) is reset after every manual retry. (`500 -> auto retry -> 401 -> refresh -> 500 -> auto retry -> 200` works with `retry: 1`)

### ⚠️ Breaking changes

- `options.retry` is no longer decremented between attempts
- the error hooks and `onResponse` may now return a `RetryIntent` (return type widened from `void`)

## Summary by CodeRabbit

* **New Features**
  * Added configurable manual and automatic request retries through request and response hooks.
  * Added retry history, including causes, triggers, status codes, and delays.
  * Added controls to trigger or cancel pending retries.
  * Retry details are now available on fetch errors.
  * Timeouts now reset for each retry attempt.

* **Bug Fixes**
  * Improved abort handling during retry delays and timeout scenarios.
  * Prevented retries for non-replayable stream request bodies.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -49,6 +49,12 @@ export function createFetchError<T = any>(
     });
   }
 
+  Object.defineProperty(fetchError, "retries", {
+    get() {
+      return ctx.retries?.history;
+    },
+  });
+
   for (const [key, refKey] of [
     ["data", "_data"],
     ["status", "status"],
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -7,15 +7,21 @@ import {
   detectResponseType,
   resolveFetchOptions,
   callHooks,
+  callRetryHooks,
+  createRetryIntent,
+  createRetryHistory,
+  mergeRetryOptions,
+  sleep,
 } from "./utils.ts";
 import type {
   CreateFetchOptions,
   FetchResponse,
-  ResponseType,
   FetchContext,
   $Fetch,
   FetchRequest,
   FetchOptions,
+  RetryEntry,
+  RetryTrigger,
 } from "./types.ts";
 
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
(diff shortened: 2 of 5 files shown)


cernymatej:
some use case examples:

**exponential backoff**

```ts
await $fetch("/flaky", {
  retry: 5,
  // or just `ctx.retries.count()` if no other retries are present
  retryDelay: (ctx) => 100 * 2 ** ctx.retries.count('auto'),
});
```

**simple polling**

```ts
await $fetch(`/jobs/${id}`, {
  onResponse(ctx) {
    if (ctx.response._data?.pending && ctx.retries.count("poll") < 10) {
      return ctx.retry({ cause: "poll", delay: 1000 });
    }
  },
});
```

**auth session refresh**

```ts
await $fetch('/protected', {
  async onResponseError(ctx) {
    if (ctx.response.status === 401 && ctx.retries.count("auth") === 0) {
      await refreshSession()
      return ctx.retry({
        cause: "auth",
        options: { headers: { /* ... */ } },
      })
    }
  },
})
```

PR #624, opened 2026-08-17
fix: normalize errors thrown while reading the response body

### LINKED ISSUE #620: `timeout` option doesn't timeout body parsing
### Environment

ofetch: 1.5.1
Node.js: v22.21.1

### Reproduction

https://github.com/dword-design/demo-ofetch-response-body-timeout

```sh
node index.js
```

### Describe the bug

Looks like the `timeout` option only covers the request phase, but when a response is incomplete, the body parsing hangs.

**Expected**
When the response is incomplete, the request is aborted with a timeout error.

**Actual**
The request doesn't return.

### Additional context

It works with `{ signal: AbortSignal.timeout(<timeout>) }`. Probably because this is passed to the underlying fetch call.

MFA-G:
Closes #620

### Problem

`timeout` is applied via `AbortSignal.timeout()` before the `fetch()` call, so on `main` it *does* already fire while the response body is still streaming (the v1 `setTimeout` + `AbortController` approach did not — that part of #620 was fixed by the `AbortSignal.timeout` refactor).

What's still broken is the **error that comes out**. Reading the body happens outside the `try`/`catch` that wraps `fetch()`, so any failure there rejects with the raw underlying error:

```
DOMException [TimeoutError]: The operation was aborted due to timeout
```

instead of the normalized `FetchError`. Practical consequences:

- `error.request`, `error.response`, `error.options` and `error.data` are all missing
- the `onRequestError` hook is never called
- the message has no `[GET] "<url>"` prefix, unlike every other failure mode

A timeout that happens 1 ms before the headers arrive and one that happens 1 ms after produce two very different error shapes today. The same applies to a connection reset mid-stream, or `JSON.parse` throwing on a truncated payload.

### Change

- `src/fetch.ts`: wrap the response-body reading `switch` in `try`/`catch` and route failures through the existing `onRequestError` + `onError()` path, so they are normalized exactly like fetch-phase errors (retry handling included).
- `src/fetch.ts`: drop the leftover `abortTimeout` variable and its `finally { clearTimeout(...) }` block — it was never assigned since the switch to `AbortSignal.timeout()`, so it was dead code.
- `test/index.test.ts`: add a `/timeout-body` route that flushes headers and then keeps the body open forever, plus a regression test asserting the request rejects (rather than hanging) with `cause.name === "TimeoutError"`.

Without the `src` change the new test fails on `cause` being `undefined`, since the raw `DOMException` has no `cause`.

### Validation

```
vitest run   → 29 passed (29)
eslint .     → clean
prettier -c  → clean
tsc --noEmit → clean
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for slow, incomplete, or streaming response bodies.
  * Requests that time out while reading responses now report a consistent `TimeoutError`.
  * Response-body failures now follow standard error and retry handling.
  * Transient body-read failures can be retried, while invalid JSON errors are not.
  * Timed-out streaming responses now abort cleanly without hanging.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -36,7 +36,10 @@ const nullBodyResponses = new Set([101, 204, 205, 304]);
 export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
   const { fetch = globalThis.fetch } = globalOptions;
 
-  async function onError(context: FetchContext): Promise<FetchResponse<any>> {
+  async function onError(
+    context: FetchContext,
+    isTransportError = false
+  ): Promise<FetchResponse<any>> {
     // Is Abort
     // If it is an active abort, it will not retry automatically.
     // https://developer.mozilla.org/en-US/docs/Web/API/DOMException#error_names
@@ -57,10 +60,10 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
       // A failure while reading the body (aborted/timed out stream, network
       // error mid-stream) is a transport failure even though the response
       // headers already arrived, so retry eligibility must not be decided from
-      // the (successful) status code of that response. Parse failures are not
-      // retryable since replaying the request cannot make the payload valid.
-      const isTransportError =
-        !!context.error && context.error.name !== "SyntaxError";
+      // the (successful) status code of that response. Failures raised while
+      // parsing an already-read body are not transport errors: the request did
+      // succeed, and replaying it (possibly non-idempotent) cannot make the
+      // payload valid.
       const responseCode =
         (!isTransportError && context.response && context.response.status) ||
         500;
@@ -197,7 +200,7 @@ export function createFetch(globalOptions: CreateFetchOptions = {}): $Fetch {
           context.options.onRequestError
         );
       }
-      return await onError(context);
+      return await onError(context, true);
     }
 
     const hasBody =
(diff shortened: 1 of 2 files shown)


MFA-G:
Thanks for the review — both points addressed in 7d35112.

**1. Retries for timeout/transport body failures (major)** — confirmed valid. After a body-read failure `context.response` is still set with its *successful* status, so `onError` computed `responseCode = 200` and the retry never fired. `context.response` is kept for the normalized error metadata, but the retry decision now treats a non-`SyntaxError` body failure as a transport error and falls back to `500` (a retryable status). Parse failures (`SyntaxError` from `JSON.parse`) stay non-retryable, since replaying the request cannot make an invalid payload valid.

Added a regression test `retries when the body read times out` using `retry: 1` + `onRequest` counter; it asserts two attempts. Verified it fails (`expected 1 to equal 2`) with the fix reverted and passes with it.

**2. Fallback timer in the streaming-timeout test (minor)** — applied the suggested `finally(() => clearTimeout(hangTimer))`, and the new test does the same.

Validation: `vitest run` 30/30 passing, `eslint .` + `prettier -c` clean, `tsc --noEmit` clean.

MFA-G:
Thanks @coderabbitai — valid, fixed in 4fba64c.

Classifying by error *name* was the weak part: a custom `parseResponse` throwing `Error` or `TypeError` looked identical to a transport failure, so `retry: 1` replayed a request whose body had already been read successfully — bad for non-idempotent requests.

Rather than guessing from the error, the two cases are now tracked at the point they can happen. Reading the body is a continuation of the transport; parsing happens after the request already succeeded:

```ts
let isTransportError = true;
try {
  switch (responseType) {
    case "json": {
      const data = await context.response.text();
      isTransportError = false; // body is fully read; anything after this is parsing
      ...
```

and `onError(context, isTransportError)` decides retry eligibility from that flag instead of `error.name`. The `SyntaxError` special case is gone — `JSON.parse` failures are covered by the same rule, and so are custom parsers regardless of what they throw.

Added `does not retry when parseResponse throws` (throwing `TypeError`, `retry: 1`), which asserts a single attempt; it fails on the previous commit and passes now.

`vitest run`: 31/31 passing. `eslint` and `tsc --noEmit` clean.

PR #635, opened 2026-09-16
fix: resolve the method from a prebuilt `Request` input

### LINKED ISSUE #632: Prebuilt write Requests bypass the default no-retry policy
## Reproduction

On main at 1dbc37fd1ceab832fc7c90cad81b1091c95ba563, passing a prebuilt Request with a write method does not apply the default no-retry policy documented for POST, PUT, PATCH, and DELETE:

```ts
import { createFetch } from "ofetch";
let calls = 0;
const ofetch = createFetch({
  fetch: async () => {
    calls++;
    return new Response("unavailable", { status: 503 });
  },
});
await ofetch(new Request("https://example.test/", { method: "POST" })).catch(() => {});
console.log(calls); // 2, expected 1
```

The retry, request-body serialization and HEAD response-body guards inspect options.method only, although native fetch also takes the method from the Request input. The same discrepancy leaves an overriding JSON object body unserialized for a POST Request and attempts body parsing for a HEAD Request with a custom fetch implementation returning a body.

Six new regression cases fail against this commit: all four write methods, HEAD parsing, and JSON body serialization. GET retry, explicit retry opt-in, and explicit method override controls pass. A local HTTP server also reproduces the POST retry behavior.

## Expected behavior

Use the effective method (resolved options first, then Request, then GET) for these three existing guards, including after an onRequest hook changes the Request. Do not change the explicit numeric retry opt-in or string-request behavior.

This is separate from #583, which changes FetchError message precedence. No error-message code needs to change for this defect.

MFA-G:
Fixes #632.

## Problem

`fetch` reads the method from the init object when it is set and otherwise from a `Request` input. Three guards in `$fetchRaw` consulted `options.method` only, so a prebuilt `Request` with a write method was treated as a `GET`:

```ts
const ofetch = createFetch({
  fetch: async () => { calls++; return new Response("unavailable", { status: 503 }) },
});
await ofetch(new Request("https://example.test/", { method: "POST" })).catch(() => {});
// calls === 2, but POST is documented as not retried by default
```

Three separate symptoms:

1. **Retry** — `POST`/`PUT`/`PATCH`/`DELETE` passed as a `Request` got the `GET` policy of one retry, so a non-idempotent request was replayed.
2. **Body serialization** — an overriding object body on a `Request` POST skipped `JSON.stringify` and the automatic `content-type: application/json`, so the raw object reached `fetch`.
3. **HEAD** — a HEAD passed as a `Request` had its response body parsed, which is observable with a custom `fetch` that returns one.

## Change

Adds `resolveMethod(request, options)` to `src/utils.ts` — `options.method` first, then `(request as Request).method`, then `GET`, upper-cased — and uses it in all three places. The resolution happens after the `onRequest` hooks so a hook that swaps the request or sets a method is honoured.

Explicit controls are deliberately untouched: a numeric `retry` and an explicit `options.method` both still take precedence over the `Request`.

## Tests

Six cases added to `test/index.test.ts`: the four write methods, HEAD parsing, and JSON body serialization cover the defect; a GET `Request`, an explicit `retry: 2`, and an explicit `method` override guard the behaviour that must not change.

The three defect tests fail on `main` with the source change reverted and pass with it. `vitest run` is 34/34, `eslint`, `prettier -c` and `tsc --noEmit` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of prebuilt requests so HTTP methods are detected consistently.
  * Retry behavior now correctly reflects the request method.
  * Request bodies are serialized and response bodies are handled correctly for POST, PUT, PATCH, DELETE, GET, and HEAD requests.
  * Explicit method and retry settings continue to take precedence when provided.
  * Improved support for URL-encoded form bodies, including content types with parameters or mixed capitalization.


MFA-G:
Addressed the CodeRabbit finding about the form media type.

`Content-Type` is a media type plus optional parameters and its casing is not significant, but the serializer compared the whole header value against the exact lowercase string. A header written as `application/x-www-form-urlencoded; charset=UTF-8` or `Application/X-WWW-Form-Urlencoded` therefore took the `JSON.stringify` path while the form media type stayed on the request, so the server got a JSON payload announced as a form. A prebuilt `Request` makes this easy to hit since it preserves whatever the constructor was given.

Extracted the check into `isFormUrlEncoded()`, which strips the parameters and lowercases the media type, and added two regression tests (parameterized + mixed case). Both fail on the previous commit and pass now; full suite is green (36 tests).

